PR #12, opened 2025-11-17
Add a PR Check workflow for branch protection

fernandol-nvidia:
## Description
- Some CI workflows are conditionally triggered based on paths changed
- This means we cannot use them directly as MR checks (because checks require that workflows are always triggered)
- Instead we will add a final "conclusion" workflow that we can use to report if an MR passes or not

## Checklist

### DIFF
--- a/.github/workflows/pr-checks.yaml
+++ b/.github/workflows/pr-checks.yaml
@@ -45,14 +45,17 @@ jobs:
           token: ''
           filters: |
             ci:
+              - '.github/workflows/pr-checks.yaml'
               - 'BUILD'
               - 'MODULE.bazel'
               - 'bzl/**'
               - 'src/**'
               - 'run/**'
             docs:
+              - '.github/workflows/pr-checks.yaml'
               - 'docs/**'
             ui:
+              - '.github/workflows/pr-checks.yaml'
               - 'BUILD'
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-checks.yaml
```diff
@@ -0,0 +1,315 @@
...
+              npm: ui/pnpm-lock.yaml
+
+      - name: Build UI
+        run: |
+          bazel build //ui/...
+
+  #######################
+  #    Status Check     #
+  #######################
+  pr-checks-status:
```
elookpotts-nvidia: I'm super confused why this is necessary - aren't skipped steps treated as success for dependent steps ([docs](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks#check-statuses-and-conclusions))? Shouldn't this be unnecessary?

fernandol-nvidia: Let me double check and see if Github branch protection can use these checks directly without the "conclusion" step.

fernandol-nvidia: You are correct, this is not necessary anymore. We can add these jobs as checks directly now that they always run.

PR #20, opened 2025-11-17
docs: update user guide links

vvnpn-nv:
## Description
update user guide links

## Checklist


docs/user_guide/workflows/interactive/port_forward.rst
ethany-nv, line 1: I think we can leave the note here, and rephrase it as:

```
For all of our examples of common services that can be port-forwarded
```

vvnpn-nv: I removed it because the link is to integration and tools folder, it's difficult for the reader to understand which one can be used for port-forward. since in the section below browser clearly tells users which one can be used I think we can remove it

PR #40, opened 2025-11-19
Adds multiversion support for docs

RyaliNvidia:
## Description
Docs now are compiled to show multiple versions for the different OSMO releases

It is currently using a forked repo of the sphinx-multiversion as sphinx-multiversion does not support building markdown

https://github.com/user-attachments/assets/3cf1f6fc-763f-45d9-9e54-8fa430dfce8c

Issue - None

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/OSMO/blob/main/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #81, opened 2025-11-23
Unify docs build into a single build, support proper multiversion links

adelbertc:
## Description 
* Unify Sphinx build to build as one project instead of separate User Guide and Deployment Guide builds
* Change hard-coded URLs to use Sphinx roles so the correct multiversion link can be interpolated at build time
* Add some overrides to `nvidia-sphinx-theme` hardcodes to get the correct sidebar Table of Contents behavior after unifying the build and by extension the `toctree`

The goal of this change is to make the links in our docs version-aware after moving to multiversion docs in https://github.com/NVIDIA/OSMO/pull/40. Normally the way to do this is to reference roles like `:doc:` and `:ref:` instead of hard-coding URLs, but since we have cross-guide links we also have to unify the Sphinx build to make the builds aware of roles across all our docs.

**NOTE**: Our root README and similar docs still hard-code to `main`. I figure this is an acceptable tradeoff at not complicating the build to account for this. All links within the actual docs pages should be cromulent.

Issue #92

## Checklist

### DIFF
--- a/docs/_templates/sidebar-nav-bs.html
+++ b/docs/_templates/sidebar-nav-bs.html
@@ -1,5 +1,5 @@
 {#
-SPDX-FileCopyrightText: Copyright (c) 2024-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 SPDX-License-Identifier: LicenseRef-NvidiaProprietary
 
 NVIDIA CORPORATION, its affiliates and licensors retain all intellectual


docs/BUILD
```diff
@@ -20,53 +20,26 @@ load("@aspect_bazel_lib//lib:copy_to_directory.bzl", "copy_to_directory")
 
 package(default_visibility = ["//visibility:public"])
 
+# Build all documentation as a unified Sphinx project
 filegroup(
-    name = "root_docs_sources",
```
adelbertc, line 24: @RyaliNvidia Can you take a look at the changes here to make sure I didn't mess anything up?

docs/Makefile
```diff
@@ -35,45 +35,21 @@ clean:
 	rm -rf _build
 
 build:
-	export OSMO_DOMAIN=public && sphinx-build -b html . -w $(ERR_DIR)/root.log $(OUT_DIR)
```
adelbertc, line 38: @RyaliNvidia Here too.

docs/_extensions/nvidia_theme_override.py
```diff
@@ -0,0 +1,33 @@
+"""Override NVIDIA theme's sidebar behavior to restore PyData defaults."""
```
RyaliNvidia, line 17: Add copyright here

docs/_templates/sidebar-nav-bs.html
```diff
@@ -0,0 +1,30 @@
+{#
+SPDX-FileCopyrightText: Copyright (c) 2024-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
RyaliNvidia: We can remove this file

docs/deployment_guide/advanced_config/rsync.rst
```diff
@@ -255,4 +255,4 @@ Troubleshooting
 
 .. seealso::
 
-  - Learn more about Rsync in OSMO at `Interactive Workflows <https://nvidia.github.io/OSMO/user_guide/main/workflows/interactive/rsync.html>`_
+  - Learn more about Rsync in OSMO at :doc:`Interactive Workflows </user_guide/workflows/interactive/rsync>`
```
RyaliNvidia: We should use ref instead of doc just so we don't need to update paths if we ever decide to the file location. General comment across the PR

docs/_templates/sidebar-nav-bs.html
adelbertc: We actually do need this change, otherwise it uses the default `sidebar-nav-bs.html` that comes with `nvidia-sphinx-theme` which hardcodes the `startdepth` and causes the guide-specific ToC's to render bad.

RyaliNvidia: NIT: 
```suggestion
SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

PR #98, opened 2025-12-01
Add a redirect to the top level docs

RyaliNvidia:
## Description
Adds a redirect to the top level docs so the url is cleaner

Issue - None

## Checklist

### DIFF
--- a/docs/_redirect/redirect_user_guide.html
+++ b/docs/_redirect/redirect_user_guide.html
@@ -2,10 +2,10 @@
 <html lang="en">
 <head>
   <meta charset="utf-8">
-  <meta http-equiv="refresh" content="0; url=release/6.0/index.html">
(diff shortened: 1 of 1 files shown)


docs/_redirect/redirect_user_guide.html
```diff
@@ -2,11 +2,10 @@
 <html lang="en">
 <head>
   <meta charset="utf-8">
-  <meta http-equiv="refresh" content="0; url=main/index.html">
-  <link rel="canonical" href="main/">
+  <meta http-equiv="refresh" content="0; url=release/6.0/index.html">
```
elookpotts-nvidia: Commenting to block auto-merge but still give approval

Do we want to redirect to 6.0 or to main? My sense is that we are linking to ToT/latest/main everywhere instead of pinning to a particular version.

elookpotts-nvidia (approved):

PR #119, opened 2025-12-09
remove pyinstaller

xutongNV:
## Description
Remove pyinstaller and use bazel to build except for cli

Issue #None

## Checklist


src/operator/BUILD
```diff
@@ -48,51 +47,37 @@ osmo_py_library(
...
+    main = "/src/operator/backend_listener.py",
+    runfiles_dir = "/src/operator/backend_listener_binary.runfiles",
+)
+
+filegroup(
+    name = "backend_listener_image_layers",
+    srcs = [
+        ":backend_listener_layer_default",
+        ":backend_listener_layer_packages",
+        ":backend_listener_wrapper",
```
fernandol-nvidia, line 79: just an FYI: the ordering of layers has implication on build cache because of how image building works under the hood.

If a previous layer changes, then subsequent layer is rebuilt (similar to how `docker build` works).

So we can order our layers from *least changing* to *most changing*.

For example, `*_default` is our application (which will prob be subjected to frequent changes).

xutongNV: will do

PR #127, opened 2025-12-10
Autogenerate Config CLI Reference Documentation

fernandol-nvidia:
## Description
- Autogenerate Config CLI Reference Docs
- Use `choices` where appropriate in `config.py`
- Removed `metavar` to prevent `choices` from rendering twice in generated docs
- Update `argparse_postprocess` to work with arbitrary levels/hierarchy
- Remove explicit program outputs from examples to prevent documentation drift as we iterate on CLIs

Issue - None

## Checklist

### DIFF
--- a/src/cli/config.py
+++ b/src/cli/config.py
@@ -877,7 +877,9 @@ List configurations in JSON format::
         help='Show a configuration or previous revision of a configuration',
         description='Show a configuration or previous revision of a configuration',
         formatter_class=argparse.RawDescriptionHelpFormatter,
-        epilog='''
+        epilog=f'''
+Available config types (CONFIG_TYPE): {CONFIG_TYPES_STRING}
+
 Examples
 ========
 
@@ -896,8 +898,8 @@ Show the ``user_workflow_limits`` workflow configuration in a previous revision:
     )
     show_parser.add_argument(
         'config',
-        choices=config_history.CONFIG_TYPES,
-        help='Config to show in format <config_type>[:<revision>]',
+        metavar='config_type',
+        help='Config to show in format <CONFIG_TYPE>[:<revision>]',
     )
     show_parser.add_argument(
         'names',
@@ -916,7 +918,9 @@ Show the ``user_workflow_limits`` workflow configuration in a previous revision:
         formatter_class=argparse.RawDescriptionHelpFormatter,
         usage='osmo config update [-h] config_type [name] [--file FILE] [--description DESCRIPTION]'
               ' [--tags TAGS [TAGS ...]]',
-        epilog='''
+        epilog=f'''
+Available config types (CONFIG_TYPE): {CONFIG_TYPES_STRING}
+
 Examples
 ========
 
@@ -936,7 +940,8 @@ Update with description and tags::
     update_parser.add_argument(
         'config',
         choices=config_history.CONFIG_TYPES,
-        help='Config type to update',
+        metavar='config_type',
+        help='Config type to update (CONFIG_TYPE)',
     )
     update_parser.add_argument(
         'name',
@@ -967,7 +972,9 @@ Update with description and tags::
         formatter_class=argparse.RawDescriptionHelpFormatter,
         usage='osmo config delete [-h] config_type [name] [--description DESCRIPTION] '
               '[--tags TAGS [TAGS ...]]',
-        epilog='''
+        epilog=f'''
+Available config types (CONFIG_TYPE): {", ".join(delete_choices)}
+
 Examples
 ========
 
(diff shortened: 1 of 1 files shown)


src/cli/config.py
```diff
@@ -858,16 +875,29 @@ def setup_parser(parser: argparse._SubParsersAction):
...
+Show the ``user_workflow_limits`` workflow configuration in a previous revision::
+
+    osmo config show WORKFLOW:3 user_workflow_limits
+'''
     )
     show_parser.add_argument(
         'config',
-        metavar='config_type',
-        help='Config to show in format <CONFIG_TYPE>[:<revision>]'
+        choices=config_history.CONFIG_TYPES,
```
elookpotts-nvidia: We want the positional argument to display as `config_type` in the help text. Can you update the name to be `config_type` so it does display as such instead of just `config`?

elookpotts-nvidia: Same for other `metavar`s that got changed.

docs/deployment_guide/references/config_cli/images/config_diff_service.png
elookpotts-nvidia, line 1: I do think that these images were helpful in conveying the intent of `osmo config diff`. Can we do one of these?

1) Somehow include the images in the rendered HTML only
2) Render the output as plain text in the epilog

(2) might be too verbose and we lose colors, so I think I lean towards (1). Your call, though, I'm flexible.

I'm ok with just one image, too, if that's easier. I would prefer the workflow diff, because that's more typical than a secret change.

src/cli/config.py
fernandol-nvidia: This is an example of `config show`. Can you share an example of what you would like to see?

```
bazel run //src/cli -- config show -h                             

usage: osmo config show [-h] {BACKEND,BACKEND_TEST,DATASET,POD_TEMPLATE,POOL,RESOURCE_VALIDATION,ROLE,SERVICE,WORKFLOW} [names ...]

Show a configuration or previous revision of a configuration

positional arguments:
  {BACKEND,BACKEND_TEST,DATASET,POD_TEMPLATE,POOL,RESOURCE_VALIDATION,ROLE,SERVICE,WORKFLOW}
                        Config to show in format <config_type>[:<revision>]
  names                 Optional names/indices to index into the config. Can be used to show a named config.

options:
  -h, --help            show this help message and exit

Examples
========

Show a service configuration in JSON format::

    osmo config show SERVICE

Show the ``default_cpu`` resource validation rule::

    osmo config show RESOURCE_VALIDATION default_cpu

Show the ``user_workflow_limits`` workflow configuration in a previous revision::

    osmo config show WORKFLOW:3 user_workflow_limits
```

docs/deployment_guide/references/config_cli/images/config_diff_service.png
fernandol-nvidia: Yes, updated the extension to:

1. strip any RST directives (e.g. `.. image::`) from the terminal output
2. keep it for docs rendering, allowing us to interweave images in the generated HTML

src/cli/config.py
elookpotts-nvidia: Ah, thanks, that's not what I was expecting; I misremembered the purpose of `metavar`. Ultimately, options are not correct here because the argument is either of the form `<CONFIG_TYPE>` e.g. `SERVICE` or of the form `<CONFIG_TYPE>:<revision>` e.g. `WORKFLOW:3`. I'm also realizing that we lost the `CONFIG_TYPE` options in the epilog. Ideally, the output would use `metavar` and include config types:

```
bazel run //src/cli -- config show -h                             

usage: osmo config show [-h] config_type [names ...]

Show a configuration or previous revision of a configuration

positional arguments:
  config_type        Config to show in format <CONFIG_TYPE>[:<revision>]
  names                 Optional names/indices to index into the config. Can be used to show a named config.

options:
  -h, --help            show this help message and exit

Available config types (CONFIG_TYPE): BACKEND, BACKEND_TEST, DATASET, POD_TEMPLATE, POOL, RESOURCE_VALIDATION, ROLE, SERVICE, WORKFLOW

Examples
========

Show a service configuration in JSON format::

    osmo config show SERVICE

Show the ``default_cpu`` resource validation rule::

    osmo config show RESOURCE_VALIDATION default_cpu

Show the ``user_workflow_limits`` workflow configuration in a previous revision::

    osmo config show WORKFLOW:3 user_workflow_limits
```

Is there a limitation preventing us from using metavar? Can we include the available config type options in the epilog?

elookpotts-nvidia: I would prefer not to include the config type options in the help string, if possible.

RyaliNvidia (approved):

src/cli/config.py
fernandol-nvidia: Done, thanks!

PR #138, opened 2025-12-11
Fix infinite refresh loop in UI when service is bad; remove Nvidia font

patclarknvidia:
## Description
Remove reference to internal Nvidia fonts
Fix infinite refresh loop in UI when service is bad

Issue - None

## Checklist

image

### DIFF
--- a/ui/src/components/AuthProvider.tsx
+++ b/ui/src/components/AuthProvider.tsx
@@ -117,7 +117,7 @@ class Auth {
       const data = await res.json();
       this.authEnabled = data.auth_enabled;
     } catch (error) {
-      console.warn(`Failed to fetch login info: ${error instanceof Error ? error.message : "Unknown error"}`);
+      throw new Error(`Failed to fetch login info: ${error instanceof Error ? error.message : "Unknown error"}`);
     }
 
     if (!this.authEnabled) {


ui/src/styles/globals.css
```diff
@@ -63,7 +63,7 @@ SPDX-License-Identifier: Apache-2.0
 
 @layer base {
   body {
-    @apply font-['Nvidia_Sans','Nvidia_Sans_Fallback'] bg-bg text-text text-sm;
```
elookpotts-nvidia, line 66: Do we need to remove the font sheet from the build? How was it being loaded?

ui/src/components/AuthProvider.tsx
```diff
@@ -257,21 +252,52 @@ class Auth {
...
+    setIsLoading(true);
+    setIsError(false);
+
+    void auth
+      .login()
+      .then(() => {
+        setIsLoading(false);
+      })
+      .catch((error) => {
+        console.warn(`Failed to login: ${error instanceof Error ? error.message : "Unknown error"}`);
```
elookpotts-nvidia: Can we both `console.warn` and display the content of the error to the user on the page in a `<pre>` tag? I think that will be easier for an admin to debug.

ui/src/components/AuthProvider.tsx
```diff
@@ -163,21 +165,19 @@ class Auth {
     try {
```
elookpotts-nvidia, line 165: Should we throw an error for the warning on L120 as well, when fetching `/auth/login_info` fails? (Apologies, GitHub won't let me comment on that line of the diff.)

ui/src/components/AuthProvider.tsx
```diff
@@ -257,21 +252,52 @@ class Auth {
...
+              window.location.reload();
+            }}
+          >
+            Reload
+          </button>
+        </PageError>
+      </div>
+    );
+  }
+
```
elookpotts-nvidia, line 305: Sweet! Do you mind adding a screenshot to the PR description?

elookpotts-nvidia: Do you mind adding repro steps to the PR description as well? I'm sure we will forget how to repro this in a couple of months 😆

ui/src/styles/globals.css
patclarknvidia: Yes - we cannot use Nvidia font as it's internal only.

ui/src/components/AuthProvider.tsx
patclarknvidia: This is a warning... and in the prod code will result in a warning and a return from login. 
Throwing will result in an error - which I think is more correct, but since you wrote the orig logic, what do you think?

patclarknvidia: Done

elookpotts-nvidia: I think it could be a warning or error, but error is definitely more obvious that something is wrong.

elookpotts-nvidia (approved):

fernandol-nvidia (approved):

PR #141, opened 2025-12-11
Support for Azure workload identity in AKS and Arc clusters

agreaves-ms:
## Description

Adds Azure AKS Workload Identity support, allowing pods to authenticate to Azure Blob Storage without connection strings or managed secrets. Uses `DefaultAzureCredential` for token-based auth and updates Helm charts to support workload identity configuration.

Related to Issue #153

### Helm charts

- all charts (service, router, backend-operator, web-ui) now support ServiceAccount configuration:
  - `serviceAccount.create` - toggle chart-managed ServiceAccount creation
  - `serviceAccount.name` - specify pre-provisioned SA name
  - `serviceAccount.annotations` - add workload identity annotations (e.g., `azure.workload.identity/client-id`)
- added `extraPodLabels` to services for workload identity labels (`azure.workload.identity/use: "true"`)
- made `imagePullSecrets` optional when using workload identity

### Azure storage backend

- `AzureBlobStorageClient` supports both connection string and token-based auth
- added `DefaultAzureCredential` support for workload identity
- SAS token generation now uses user delegation keys instead of account keys
- credential fields (`access_key_id`, `access_key`) are now optional

### Credentials

- `BasicDataCredential` fields are now optional
- added `get_access_key_value()` for safe credential retrieval
- updated callers to use new accessor pattern

### PyInstaller

- added `azure.identity`, `msal`, and `msal_extensions` to Azure hooks for CLI workload identity support

### Tests

- added test coverage for workload identity auth paths

## Usage

```yaml
serviceAccount:
  create: false
  name: "my-workload-identity-sa"
  annotations:
    azure.workload.identity/client-id: "<client-id>"

services:
  service:
    extraPodLabels:
      azure.workload.identity/use: "true"
```

## Checklist

### DIFF
--- a/src/lib/data/storage/backends/azure.py
+++ b/src/lib/data/storage/backends/azure.py
@@ -287,7 +287,6 @@ def _extract_account_key_from_connection_string(connection_string: str) -> str:
 
 def create_client(
     data_cred: credentials.DataCredential,
-    storage_account: str | None = None,  # pylint: disable=unused-argument
     account_url: str | None = None,
 ) -> blob.BlobServiceClient:
     """
@@ -880,7 +879,6 @@ class AzureBlobStorageClientFactory(provider.StorageClientFactory):
     """
 
     data_cred: credentials.DataCredential
-    storage_account: str
     account_url: str
 
     @override
@@ -888,7 +886,6 @@ class AzureBlobStorageClientFactory(provider.StorageClientFactory):
         return AzureBlobStorageClient(
             azure_client_factory=lambda: create_client(
                 self.data_cred,
-                storage_account=self.storage_account,
                 account_url=self.account_url,
             ),
             data_cred=self.data_cred,
--- a/src/lib/data/storage/backends/backends.py
+++ b/src/lib/data/storage/backends/backends.py
@@ -910,7 +910,6 @@ class AzureBlobStorageBackend(common.StorageBackend):
         def _validate_auth():
             with azure.create_client(
                 data_cred,
-                storage_account=self.storage_account,
                 account_url=self.auth_endpoint,
             ) as service_client:
                 if self.container:
@@ -960,7 +959,6 @@ class AzureBlobStorageBackend(common.StorageBackend):
 
         return azure.AzureBlobStorageClientFactory(
             data_cred=data_cred,
-            storage_account=self.storage_account,
             account_url=self.auth_endpoint,
         )
 
--- a/src/lib/data/storage/backends/tests/test_backends.py
+++ b/src/lib/data/storage/backends/tests/test_backends.py
@@ -247,7 +247,6 @@ class AzureDefaultDataCredentialTest(unittest.TestCase):
         # Act
         azure.create_client(
             data_cred,
-            storage_account='mystorageaccount',
             account_url='https://mystorageaccount.blob.core.windows.net',
         )
 


fernandol-nvidia:
This is awesome, @agreaves-ms! Thank you so much for looking into this.

The team will be going through this change and thinking through what else we might need to change. Some high level thoughts is we probably need to give you more "types" of Credentials (e.g. WorkloadIdentityCredential, DataAccessCredential, ConnectionStringCredential, etc). 

Another thing we will look at as a team is changes necessary to our service and `osmo-ctrl` to support this. Please stay tuned! And again, really appreciate the great work so far!

fernandol-nvidia:
Hey @agreaves-ms, thanks for the update and apologies for the delay. I spent the past couple of days doing a refactor of how OSMO data operations work w.r.t to different credential strategies. I have a pending PR at https://github.com/NVIDIA/OSMO/pull/159.

The tl;dr is that we are differentiating `DataCredential` into `StaticDataCredential` (what we currently have) and `DefaultDataCredential` (which we delegate to the underlying SDK on credential resolution).

I believe this would give you the necessary abstractions/interface to add workload identity support for Azure. More specifically:

Defining the behavior at https://github.com/NVIDIA/OSMO/blob/682dbf350ceb6d5ada286cf16fbaf077943ac0eb/src/lib/data/storage/backends/azure.py#L274-L287

And overriding this method on how to decide if workload identity should be available for a particular backend:
https://github.com/NVIDIA/OSMO/blob/682dbf350ceb6d5ada286cf16fbaf077943ac0eb/src/lib/data/storage/backends/common.py#L279-L297

Hopefully, this frees you from having to update so many unrelated files just to get Azure workload identity working :)

Please feel free to add any feedback/suggestions that would be helpful for you. I will keep you updated as things progress and can help you rebase this PR to absorb the new changes.

agreaves-ms:
@fernandol-nvidia I've update this PR with the changes that have since been added. I needed to make some additional changes to parts that would allow for DataCredential instead of StaticDataCredential (only) as well.

Please review, I built this and verified all the azure workload identity, RBAC, and local auth works with these changes. Feel free to take over this PR as well as needed.

src/utils/connectors/postgres.py
fernandol-nvidia, line 1: We can leave this file unchanged because non-staticdatacredentials doesn't need to be persisted into DB.

src/utils/job/task.py
fernandol-nvidia, line 1: We can leave this file unchanged for similar reasons. Non-staticdatacredential won't be encoded in the task.

src/lib/data/storage/backends/azure.py
```diff
@@ -271,6 +274,35 @@ def __next__(self) -> bytes:
...
+    Connection strings use semicolon-delimited key=value format:
+    DefaultEndpointsProtocol=https;AccountName=...;AccountKey=...;EndpointSuffix=...
+    """
+    for part in connection_string.split(';'):
+        if part.startswith('AccountKey='):
+            return part[len('AccountKey='):]
+    raise ValueError('AccountKey not found in connection string')
+
+
 def create_client(data_cred: credentials.DataCredential) -> blob.BlobServiceClient:
```
fernandol-nvidia: We can pass in the params (`storage_account`, `account_url`) from the callers (instead of deriving it here). This is because the callers (i.e. an `AzureBlobStorageBackend`) will have:

1. `storage_account` (https://github.com/NVIDIA/OSMO/blob/main/src/lib/data/storage/backends/backends.py#L822-L825)
2. `endpoint`
https://github.com/NVIDIA/OSMO/blob/main/src/lib/data/storage/backends/common.py#L164-L168
https://github.com/NVIDIA/OSMO/blob/main/src/lib/data/storage/backends/backends.py#L865-L868

The callsites of `create_client` are:

https://github.com/NVIDIA/OSMO/blob/main/src/lib/data/storage/backends/backends.py#L911

and

https://github.com/NVIDIA/OSMO/blob/main/src/lib/data/storage/backends/azure.py#L838.

For the second one, we can add the necessary params to `AzureBlobStorageClientFactory` from (https://github.com/NVIDIA/OSMO/blob/94b1827bce5288b3fa096866efb572dfb1e0fcc2/src/lib/data/storage/backends/backends.py#L957)

fernandol-nvidia (commented):
Happy 2026 :)

Nice, this looks great! I've added some minor comments which will help slim down the changes necessary.

Given that you've already tested this, I feel pretty confident with the PR. Once comments are addressed, I will approve!

src/utils/connectors/postgres.py
agreaves-ms: I've been trying to figure why I needed this change as I was getting an error here when I was testing it, same thing with the LogConfig. I believe my workflow config was still stored in my database with the access key id, as I've been testing without this change and after I fully destroyed and redeployed my config, it doesnt seem to have the problem anymore.

I've gone ahead and reverted these changes.

src/lib/data/storage/backends/azure.py
agreaves-ms: Great call, it looks like I only needed `account_url` as `storage_account` wasn't being used. Hopefully the changes here are what you're expecting as well.

Issue #147, opened 2025-12-12
Project: Backend Operators Redesign


xutongNV:
### Have you read the Project Process docs?

- [x] Yes, I have read and understood the RFC docs

### Summary

This project aims to redesign the OSMO backend operator component to address critical scaling, reliability, and performance issues.

### Person in Charge (PIC)

@xutongNV

### Motivation

To prepare for future growth, OSMO's backend operators require architectural enhancements to support production workloads for Kubernetes backends at large scale. The current system provides a solid foundation, but can be optimized in several key areas: system stability under sustained load, event delivery consistency, and workflow status update latency. 

This redesign will proactively strengthen OSMO's ability to support production-scale deployments reliably and efficiently as customer workloads continue to grow. Rewriting the backend listener using Golang will allow us to leverage native Kubernetes Go library for more performant operations and more built-in features (such as node/pod events caching).

### Problem

1. Scaling - Single-threaded listener design limits workflow throughput
2. Reliability - Frequently restarts, and event drops
3. Performance - Memory leaks consume several GB per instance, and workflow status updates experience high latency

PR #149, opened 2025-12-12
#147 - backend operator redesign doc

xutongNV:
## Description

Issue #147 

## Checklist

### DIFF
--- a/projects/PROJ-147-operator-redesign/benchmark.md
+++ b/projects/PROJ-147-operator-redesign/benchmark.md
@@ -125,6 +125,7 @@ METRIC 2: CROSS-IMPLEMENTATION DETECTION LATENCY
     Python detected first: 0 times
 
   Time Difference Statistics:
+  (For Go it takes around 20 ms from creating an Message to getting an ACK)
     Average:           29.42ms
     Median:            24.48ms
     Min:                4.76ms
@@ -144,8 +145,6 @@ METRIC 2: CROSS-IMPLEMENTATION DETECTION LATENCY
 ================================================================================
 UNIQUE POD EVENT DETECTION ANALYSIS
 ================================================================================
-
-Benchmark Duration:  Python: 64.4s  |  Go: 62.6s
 
 EVENT COVERAGE:
   Total unique events:     951


projects/PROJ-147-operator-redesign/PROJ-147-operator-redesign.md
```diff
@@ -0,0 +1,295 @@
...
+
+**Observability**
+- Lack of proper KPIs and logging infrastructure
+- Insufficient test coverage makes it difficult to diagnose issues and measure system health
+
+## Use Cases
+
+| Use Case | Description |
+|---|---|
+| Large comupting backend | Backend operator needs to be able to handle a large computing backends, e.g. a k8s cluster with 2k nodes |
```
ecolternv: NIT: comupting -> computing

projects/PROJ-147-operator-redesign/PROJ-147-operator-redesign.md
```diff
@@ -0,0 +1,295 @@
...
+- **Event Listener**: Listens to scheduler events
+- **Operator (renamed from Agent) Service**: Receives backend messages, puts them in the message queue and acknoledges
+- **Message Queue (Redis)**: Message broker for compacting and asynchronous processing
+- **Worker**: Task executor that processes messages from the queue
+
+## Detailed Design
+
+### Decouple Backend Listener (Golang)
+
+(TODO) decide multiple executables or just coroutines
```
ecolternv: One executable with mulitple corroutines should be enough for most cases, especially since golang can take advantage of multiple CPU cores. 
Maybe we could offer both options. By default the executable will run all 3 listeners 
```
backend-listener
```
But with command line args, you can make it only run one listener, so you could run multiple copies of it
```
backend-listener --workflow-listener
backend-listener --resource-listener
backend-listener --event-listener
```
So if you want them separate for a larger cluster, you can do that

projects/PROJ-147-operator-redesign/PROJ-147-operator-redesign.md
```diff
@@ -0,0 +1,295 @@
...
+
+The proposed architecture divided Backend Listener into 3 aspects: Updating OSMO pod status (most important), aggregating resource usages and stream scheduling events.
+
+Decoupling Backend Listener into separate Go-based services leverages Go's goroutines to achieve true parallelism and higher CPU utilization, simplifies management by giving each listener a single well-defined responsibility, and most importantly, gives workflow status updates prioritization so that they are no longer blocked by high-volume resource/node messages.
+
+### Deprecate Backend Heartbeat
+
+Currently Backend Listener sends heartbeats following a schedule to Agent Service. Agent Service updates Database with `last_heatbeat` to determine whether a backend is marked as `online`. Based on the following considerations the new resign will deprecate the heartbeat message and backend online status:
+
+- Heartbeat is not reliablly sent and its frequency requires tunning for different compute backend sizes
```
ecolternv: >Heartbeat is not reliablly sent
This seems like a bug in the way we have implemented it, we should be able to implement it to be reliably sent

>its frequency requires tunning for different compute backend sizes
What does backend size have to do with frequency of heartbeat? Shouldn't we alway send it at the same frequency?

xutongNV: These two are the caveats for the current implementation. Because backend listener and agent have to deal with other messages in coroutines which are often in a burst, the heartbeat message can be buried and delayed in sending and processing. This make the effective frequency a lot smaller that what we set in the cmd argument. For gb200 cluster we have to make the frequency very high so that the real frequency lands on the service side is within expected range

xutongNV: I'll modify this line in readme

projects/PROJ-147-operator-redesign/benchmark.md
```diff
@@ -0,0 +1,375 @@
...
+## k8s Library - Python vs Go
+
+### Overview
+
+This section presents comprehensive performance benchmarks comparing Python and Go k8s libraries with cache implementation.
+
+### Startup Perfomance
+
+**Test Configuration:**
+- 85 Pods exist in **osmo-prod** namespace in Isaac cluster with no status changes
```
RyaliNvidia: Please remove the cluster information here

projects/PROJ-147-operator-redesign/benchmark.md
```diff
@@ -0,0 +1,375 @@
...
+- Service receives the message and ACKs immediately
+
+**Benchmark Results**
+
+```
+================================================================================
+UNIQUE POD EVENT DETECTION ANALYSIS
+================================================================================
+
+Benchmark Duration:  Python: 64.4s  |  Go: 62.6s
```
ecolternv: Im a little confused on how this relates to `Average difference` My understanding of this is that python took 2 seconds more than golang to send all 951 pod events. But the average difference is 8s, so I would expect that python was 8s behind and should have taken about 8s longer than golang. Unless maybe python was behind for most events and then caught up near the end?

projects/PROJ-147-operator-redesign/benchmark.md
```diff
@@ -0,0 +1,375 @@
...
+    Python pods with transitions: 55
+    Go pods with transitions:     55
+    Matched transitions analyzed: 55
+
+  Detection Statistics:
+    Go detected first:     55 times
+    Python detected first: 0 times
+
+  Time Difference Statistics:
+    Average:           29.42ms
```
ecolternv, line 129: Do we have the absolute time golang took and that python took?

1000ms for golang and 1029.42ms for python doesn't matter much
1ms for golang and 30.42ms for python is huge

projects/PROJ-147-operator-redesign/benchmark.md
```diff
@@ -0,0 +1,375 @@
...
+Event 10:    1,154ms   ← Over 1 second
+Event 20-:   2,240ms   ← Stabilized at max queue depth
+```
+
+#### Go Implementation Results (1 Worker)
+
+Processing Time Statistics:
+```
+Min:     144.19ms   (slower cold start than Python)
+Max:     1,135.00ms ✅ (50% better than Python peak!)
```
ecolternv, line 301: Very nice to see that golang with one worker is faster than python, but I wonder where this speedup is coming from, since the actually processing of the event is happening in python still. Is the extra slowness from the python backend operator being slower than the golang backend operator?

projects/PROJ-147-operator-redesign/benchmark.md
xutongNV: I didnt record the absolute time because I cant get the true time of when an event happen, I can only know when python or go detect it.

projects/PROJ-147-operator-redesign/benchmark.md
xutongNV: I can remove this if it is misleading. Basically my script launches python server, python listener, go server and go listener as four commands so there's small gap between commands. Go pairs are ready 2 seconds later than python and they run for about one minute

projects/PROJ-147-operator-redesign/benchmark.md
xutongNV: there are multiple reasons, for listener go can send all events almost in parallel, grpc is faster then websocket. But I think most importantly, go has the worker as a whole new process but python has to deal with context switch between receiving coroutine, managing memory message queue and processing coroutine.

projects/PROJ-147-operator-redesign/benchmark.md
xutongNV: round time for Go to create and send the message and receive ACK is about 20 ms. But I cant measure how long it takes between pod status change and Go detect that change.

In real case, the difference will be trivial if pods are stable and status wont change for hours. But this could be important if there's a burst in change when submitting workflow with many tasks

projects/PROJ-147-operator-redesign/benchmark.md
ecolternv: Oh, so each one runs for 1 minute after it finishes starting? Yeah, maybe we can leave this out then ,or just mark in the test description that the services run for 1 minute. I dont think this exact timing adds much info and its just kind of confusing

projects/PROJ-147-operator-redesign/benchmark.md
ecolternv: If its too hard to get the exact timing thats ok as long as we know its about 20ms for golang. That helps quantify that python is over 2x slower than golang, so there is a significant benefit to switching.

ecolternv (approved):

RyaliNvidia (approved):

elookpotts-nvidia (approved):

PR #151, opened 2025-12-13
Allow admins to update an invalid backend type

RyaliNvidia:
## Description
Right now, if we have an older supported backend type set, the service will not be able to start up. This fixes so we can start and update it to a valid type

Issue #None

## Checklist

### DIFF
--- a/src/utils/connectors/postgres.py
+++ b/src/utils/connectors/postgres.py
@@ -2139,6 +2139,7 @@ class BackendSchedulerSettings(pydantic.BaseModel):
     """Settings that control the how pods are scheduled in a backend"""
     scheduler_type: BackendSchedulerType = BackendSchedulerType.KAI
     scheduler_name: str = 'kai-scheduler'
+    scheduler_timeout: int = 30
 
 
 class BackendNodeConditions(pydantic.BaseModel):


src/utils/connectors/postgres.py
```diff
@@ -4063,6 +4080,7 @@ def has_access(self, path: str, method: str) -> bool:
         policies=[
             role.RolePolicy(
                 actions=[
+                    role.RoleAction(base='http', path='/api/auth/access_tokenq', method='*'),
```
xutongNV: remove

RyaliNvidia: This should have been removed in another PR

src/utils/connectors/postgres.py
```diff
@@ -2125,15 +2132,25 @@ def list_from_db(cls, backends: List[str] | None = None,
 
 class BackendSchedulerType(enum.Enum):
     """ Defines the type of scheduler used by the backend """
+    DEFAULT = 'default'
```
xutongNV: are we adding this back?

src/utils/connectors/postgres.py
RyaliNvidia: Whoa. That is strange how this got added. I'll remove it

src/utils/connectors/postgres.py
RyaliNvidia: no. Issue with stashing

vvnpn-nv (approved):

Issue #153, opened 2025-12-15
Support different data credential resolution strategies for Data Storage Access


fernandol-nvidia:
### What new feature or improvement would you like to see?

OSMO System to be able to use various data credential resolution strategies:

Current Support:
- File system/config (this is currently supported)

Immediate future:
- Workload Identity on CSPs (not supported)

Distant future:
- Environment Variable (not supported)
- Vault (not supported)

### Why is this needed?

Allow OSMO service + workflows to be able to leverage CSP's workload identity (e.g. Azure, GCP, AWS) without explicit credentials to be configured and passed around.

### Code of Conduct

- [x] I agree to follow NVIDIA OSMO's Code of Conduct
- [x] I have searched the [open feature requests](https://github.com/NVIDIA/OSMO/issues?q=is%3Aopen+is%3Aissue+label%3Afeature) and have found no duplicates for this feature request

fernandol-nvidia:
Implementation is complete. We now have ways to:

1. Specify whether a storage backend is capable of credential-less operations
2. S3 already supports credential-less out of the box
3. Client-side validation (because the OSMO service won't necessarily be in the same cluster/have the same workload identity), so we need to do client-side validation inside osmo `ctrl`

PR #159, opened 2025-12-16
Support Credential-less Data Operations in Data Access Layer

fernandol-nvidia:
## Description

Issue #153

## Checklist

### DIFF
--- a/src/lib/data/storage/backends/backends.py
+++ b/src/lib/data/storage/backends/backends.py
@@ -144,23 +144,14 @@ class Boto3Backend(common.StorageBackend):
         if data_cred is None:
             data_cred = self.resolved_data_credential
 
-        match data_cred:
-            case credentials.StaticDataCredential():
-                return s3.S3StorageClientFactory(  # pylint: disable=unexpected-keyword-arg
-                    data_cred=data_cred,
-                    region=region,
-                    scheme=self.scheme,
-                    endpoint_url=self.auth_endpoint if self.auth_endpoint else None,
-                    extra_headers=self._get_extra_headers(request_headers),
-                    supports_batch_delete=self.supports_batch_delete,
-                )
-
-            case credentials.DefaultDataCredential():
-                raise NotImplementedError(
-                    'Default data credentials are not supported yet')
-
-            case _ as unreachable:
-                assert_never(unreachable)
+        return s3.S3StorageClientFactory(  # pylint: disable=unexpected-keyword-arg
+            data_cred=data_cred,
+            region=region,
+            scheme=self.scheme,
+            endpoint_url=self.auth_endpoint if self.auth_endpoint else None,
+            extra_headers=self._get_extra_headers(request_headers),
+            supports_batch_delete=self.supports_batch_delete,
+        )
 
 
 class SwiftBackend(Boto3Backend):


src/service/core/data/data_service.py
```diff
@@ -50,11 +50,14 @@ def validate_user_cred(postgres: connectors.PostgresConnector, user: str, locati
     Validates that the user has the specifc access type to a backend location
     """
     backend_info = storage.construct_storage_backend(location)
+
+    # TODO: do we support workload identity credentials here?
```
fernandol-nvidia: Question for reviewers, should we support workload identity here?

Current state is no... 

Current user cases are from Dataset CLI (afaik) which won't have workload identity. So I am leaning towards not supporting it here.

src/utils/job/workflow.py
```diff
@@ -647,15 +647,31 @@ def _fetch_bucket_info(dataset_info: common.DatasetStructure)\
...
+
+                # Check if user credentials have access to WRITE
+                try:
+                    if not is_input and bucket_info.uri not in seen_uri_output:
+                        bucket_info.data_auth(data_cred, storage.AccessType.WRITE)
+                        seen_uri_output.add(bucket_info.uri)
+                except osmo_errors.OSMOCredentialError as err:
+                    has_access = False
+            else:
+                # TODO: check pod template to see if we have workload identity configured for
```
fernandol-nvidia: @RyaliNvidia, we briefly discussed how we might be able to leverage pod template to decide if a pool can support workload identity.

Do you think this is the right place for that check? I will have a follow up PR to implement this.

src/service/core/data/data_service.py
RyaliNvidia: I feel like we should be able to support workload identity for datasets. Workflows might decide to upload to datasets using their workload identity and the service should be able to support workload identity using the service account

RyaliNvidia: I think we should operate similarly with the CLI where if there is a cred, we use that, otherwise, we check for workload identity. I guess we need to be explicit that if we want workflow datasets to use workload identity, we need the service account in the core service to have the access as well

src/utils/job/workflow.py
RyaliNvidia: So the pod template is formed at the very end of the task pod spec creation so it will be a little funky to generate it here, but even if you do, you would only be able to see if the SA has an annotation, but not if the annotation is valid.

fernandol-nvidia: Spoke with @RyaliNvidia a bit today about this. The conclusion is that despite having pod template checking, it is not reliable and low ROI.

Below are scenarios we can reliably validate data auth pre-workflow submission:

- If the service and workflow pods are deployed in the same CSP with the same workload identity
- If we know for certain certain storage backends do not support workload identity

However, if workflow runs with a different CSP/cluster/workload identity, then the only certain way to validate `data_auth` is on the workflow side .

This brings us to a follow up PR, where we will introduce data auth validation in `osmo-ctrl` prior to task input download and user task execution. This way, we can still fail reasonably fast without significant engineering efforts.

src/service/core/data/data_service.py
fernandol-nvidia: Discussed with @RyaliNvidia in depth about this as well. Workload Identity can only be reliably validated from the client. So we will add dataset auth validation on the client side. 

In the near future, @RyaliNvidia's work on auth redesign will also cover role-based authorization of dataset bucket operation.

src/utils/job/workflow.py
```diff
@@ -647,15 +647,31 @@ def _fetch_bucket_info(dataset_info: common.DatasetStructure)\
...
+
+                    # Check if user credentials have access to WRITE
+                    if not is_input and bucket_info.uri not in seen_uri_output:
+                        bucket_info.data_auth(data_cred, storage.AccessType.WRITE)
+                        seen_uri_output.add(bucket_info.uri)
+
+                except osmo_errors.OSMOCredentialError as err:
+                    # If user's existing credentials do not have access,
+                    # check if the backend supports environment authentication
+                    has_access = bucket_info.supports_environment_auth
```
RyaliNvidia: This is interesting. If the user has creds but they are bad, we check the environment auth. However, when the CLI tries to do a data operation, if there is a cred, it will never fallback to environment auth. Seems like this can cause issues right?

RyaliNvidia: I believe we don't fallback to environment auth unless I'm mistaken that this change is made in this PR

fernandol-nvidia: Hmm that's a good point. In CLI, we do not fall back to other credentials if there's an existing one (regardless of whether or not it has access). This is consistent with how SDK credential resolution works as well (i.e. if AWS environment variables are set, it won't fallback to workload identity). 

So you are correct, we should be consistent in our validation as well, if a user's explicit credentials won't work, we should just fail fast.

Great catch!

RyaliNvidia: The way that you have it currently designed is such that the uri doesn't get added to the seen list if the environment auth is there. This causes you to check every time. I guess this matters less now that you don't fallback to environment if cred fails. It could maybe go like:
```
            # Check if user credentials have access to READ
            if is_input and bucket_info.uri not in seen_uri_input:
                if data_cred is not None:
                    bucket_info.data_auth(data_cred, storage.AccessType.READ)
                else:
                    if not bucket_info.supports_environment_auth:
                        raise osmo_errors.OSMOCredentialError(
                            f'Could not validate access to {bucket_info.uri} for user {user}.')
                seen_uri_input.add(bucket_info.uri)

            # Check if user credentials have access to WRITE
            if not is_input and bucket_info.uri not in seen_uri_output:
                if data_cred is not None:
                    bucket_info.data_auth(data_cred, storage.AccessType.WRITE)
                else:
                    if not bucket_info.supports_environment_auth:
                        raise osmo_errors.OSMOCredentialError(
                            f'Could not validate access to {bucket_info.uri} for user {user}.')
                seen_uri_output.add(bucket_info.uri)
```

RyaliNvidia: This more mattered with the previous logic so this is more of a NIT

fernandol-nvidia: Updated this block to be more intuitive (and to flatten the nesting)

RyaliNvidia (approved):

vvnpn-nv (approved):

Issue #148, opened 2025-12-12
New Project: Auth Redesign


RyaliNvidia:
### Have you read the Project Process docs?

- [x] Yes, I have read and understood the RFC docs

### Summary

There are many issues plaguing the current auth design such as the tight linking between the auth middleware and the service logic itself, reliance on linking role policies to direct apis, and the dependency on Keycloak. This project will have three phases to resolve each one.

### Person in Charge (PIC)

@RyaliNvidia 

### Motivation

Currently, OSMO struggles to seamlessly integrate into different ISPs because of our reliance on Keycloak and the large configuration overhead to enable roles. By removing the Keycloak, we can rely on just the ISP and OSMO to determine user authorization as well as bring in the role mapping configuration inside of OSMO to better support access tokens.

Since role policies directly map to APIs, any changes or additions to APIs lead to very unclear behavior to admins on how to properly update old policies to match. Furthermore, the strict mapping to APIs doesn't check for more detailed authorization that is determined by the body or parameter information that is needed for authorization access to pool spec/log/exec/port-forwarding APIs. 

One more issue with the current design is the intermingling of the authorization middleware with the service. From a performance standpoint, we have not noticed any performance issues with reaching out to Postgres, but we should be very mindful of the strain this could place as every API call is making a request to postgres to check on the role information. Moreover, the fact that authorization happens after rate limiting can result in unauthorized users impacting the experience for authorized users.

### Problem

1. Improves performance for authorization validation and decouples logic from service
2. Cleaning up the roles mapping and future proofing the role policies
3. Streamlines integration with ISPs with the removal of Keycloak

PR #165, opened 2025-12-18
#148 - Auth Project Design Documents

RyaliNvidia:
## Description
Design doc for auth sidecar, idp integration, and resource action model

The numbers for the auth sidecar was from https://github.com/NVIDIA/OSMO/pull/126

Issue #148

## Checklist

### DIFF
--- a/projects/PROJ-148-auth-rework/PROJ-148-resource-action-model.md
+++ b/projects/PROJ-148-auth-rework/PROJ-148-resource-action-model.md
@@ -1,5 +1,5 @@
 <!--
-SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -826,3 +826,124 @@ var ActionRegistry = map[string][]EndpointPattern{
 6. **Test in development** — Validate all existing access patterns
 7. **Roll out to staging** — Enable feature flag and monitor
 8. **Document** — User-facing documentation for new policy format
+
+---
+
+## Open Questions
+
+### Should we have a way to define whether is pool is public/private for logs/spec/exec/port-fowarding outside of roles?
+
+Problem statement:
+
+In the current design above, roles are mapped to a group of policies and evaluated independently.
+This means that if one role evaluates to true, the authorization allows the user even in another
+role denies.
+
+We have the case where we want individual pools to restrict access so people without
+access to submit to the pool cannot view logs/view spec/exec/port-foward. However, with the design,
+we would then need to add the policy which denies access to that pool to all roles which allow access
+to the pool such as the default role.
+
+This cumbersome change will have to be done by an admin since we don't want to give users the access.
+
+Ideally, the OSMO admin should be out of the loop and it is just the pool "admin" to determine the state of the
+pool. In the future, OSMO admins should no longer need to deal with creating pools and everything
+regarding the pool can be handed over to a team member using the pool.
+
+A diagram showing the role policy model and how roles are evaluated is shown below as a reminder.
+
+```
+┌─────────────────────────────────────────────────────────────────────────────┐
+│                    ROLE-POLICY RELATIONSHIP MODEL                           │
+└─────────────────────────────────────────────────────────────────────────────┘
+
+┌──────────┐
+│   USER   │
+└────┬─────┘
+     │
+     │ has many (1:N)
+     │
+     ├─────────────┬─────────────┬─────────────┐
+     ▼             ▼             ▼             ▼
+┌─────────┐   ┌─────────┐   ┌─────────┐   ┌─────────┐
+│  Role 1 │   │  Role 2 │   │  Role 3 │   │  Role N │
+│ (admin) │   │ (user)  │   │ (custom)│   │   ...   │
+└────┬────┘   └────┬────┘   └────┬────┘   └────┬────┘
+     │             │             │             │
+     │ has one     │ has one     │ has one     │ has one
+     │ (1:1)       │ (1:1)       │ (1:1)       │ (1:1)
+     │             │             │             │
+     ▼             ▼             ▼             ▼
+┌─────────┐   ┌─────────┐   ┌─────────┐   ┌─────────┐
+│ Policy1 │   │ Policy2 │   │ Policy3 │   │ PolicyN │
+└────┬────┘   └────┬────┘   └────┬────┘   └────┬────┘
+     │             │             │             │
+     │ contains    │ contains    │ contains    │ contains
+     │ many        │ many        │ many        │ many
+     │ (1:N)       │ (1:N)       │ (1:N)       │ (1:N)
+     │             │             │             │
+     ├──────────┐  ├──────────┐  ├──────────┐  ├──────────┐
+     ▼          ▼  ▼          ▼  ▼          ▼  ▼          ▼
+┌──────────┐ ┌──┐ ┌──────────┐ ┌──┐ ┌──────────┐ ┌──┐ ┌──────────┐ ┌──┐
+│Statement1│ │..│ │Statement1│ │..│ │Statement1│ │..│ │Statement1│ │..│
+├──────────┤ └──┘ ├──────────┤ └──┘ ├──────────┤ └──┘ ├──────────┤ └──┘
+│Effect:   │      │Effect:   │      │Effect:   │      │Effect:   │
+│  Allow   │      │  Allow   │      │  Deny    │      │  Allow   │
+│Actions:  │      │Actions:  │      │Actions:  │      │Actions:  │
+│  [...]   │      │  [...]   │      │  [...]   │      │  [...]   │
+│Resources:│      │Resources:│      │Resources:│      │Resources:│
+│  [...]   │      │  [...]   │      │  [...]   │      │  [...]   │
+└──────────┘      └──────────┘      └──────────┘      └──────────┘
+
+
+┌─────────────────────────────────────────────────────────────────────────────┐
+│                        AUTHORIZATION EVALUATION                              │
+└─────────────────────────────────────────────────────────────────────────────┘
+
+  Request: POST /api/workflow/abc123/cancel
+      │
+      ▼
+  ┌────────────────────────────────────────────────────────────┐
+  │ Step 1: Resolve action from path                           │
+  │   POST /api/workflow/abc123/cancel → workflow:Cancel       │
+  └────────────────┬───────────────────────────────────────────┘
+                   │
+                   ▼
+  ┌────────────────────────────────────────────────────────────┐
+  │ Step 2: Collect ALL policies from ALL user's roles         │
+  │   User has: [Role1, Role2, Role3, ..., RoleN]              │
+  │   Collect: [Policy1, Policy2, Policy3, ..., PolicyN]       │
+  └────────────────┬───────────────────────────────────────────┘
+                   │
+                   ▼
+  ┌────────────────────────────────────────────────────────────┐
+  │ Step 3: Evaluate ALL statements from ALL policies          │
+  │                                                            │
+  │   Phase 1: Check for ANY Deny                              │
+  │   ┌─────────────────────────────────────────────┐          │
+  │   │ For each Policy (1..N):                     │          │
+  │   │   For each Statement in Policy:             │          │
+  │   │     If Effect == "Deny" AND                 │          │
+  │   │        action matches workflow:Cancel AND   │          │
+  │   │        resource matches workflow/abc123     │          │
+  │   │     → RETURN DENY (deny always wins)        │          │
+  │   └─────────────────────────────────────────────┘          │
+  │                                                            │
+  │   Phase 2: Check for ANY Allow                             │
+  │   ┌─────────────────────────────────────────────┐          │
+  │   │ For each Policy (1..N):                     │          │
+  │   │   For each Statement in Policy:             │          │
+  │   │     If Effect == "Allow" AND                │          │
+  │   │        action matches workflow:Cancel AND   │          │
+  │   │        resource matches workflow/abc123     │          │
+  │   │     → Mark as allowed                       │          │
+  │   └─────────────────────────────────────────────┘          │
+  │                                                            │
+  │   If ANY Allow found → RETURN ALLOW                        │
+  │   If NO Allow found → RETURN DENY (implicit deny)          │
+  └────────────────┬───────────────────────────────────────────┘
+                   │
+                   ▼
+            ┌──────────┐
+            │  RESULT  │
+            └──────────┘


projects/auth/PROJ-148-auth-sidecar.md
```diff
@@ -0,0 +1,548 @@
...
+## Overview
+
+The Authorization Sidecar project implements a high-performance Golang gRPC service that provides
+centralized role-based access control (RBAC) for all OSMO services through Envoy's External
+Authorization API. This replaces the Python `AccessControlMiddleware` with a more efficient,
+scalable, and maintainable solution.
+
+### Motivation
+
+- **Performance**: Python middleware adds latency to every request; Golang gRPC with caching achieves <5ms authorization checks
```
ecolternv: This isnt a strong argument for the golang sidecar. We could add caching to the python middleware too

projects/auth/PROJ-148-auth-sidecar.md
```diff
@@ -0,0 +1,548 @@
...
+- **Scalability**: Sidecar pattern scales automatically with service pods; no centralized authorization bottleneck. Rate-limiting is now be applied with authorized users.
```
ecolternv: >Sidecar pattern scales automatically with service pods
The middleware also scales automatically with the number of services pods. Each service pod will have its own copy of the middleware

projects/PROJ-148-auth-rework/PROJ-148-auth-sidecar.md
```diff
@@ -0,0 +1,548 @@
...
+| Unauthorized access attempt | User without proper roles attempts `/api/workflow` - authz_sidecar denies with 403 Forbidden |
+| Policy cache hit | Repeated requests for same role combination served from cache in <5ms |
+| Policy update | Database role policies updated - reflected in authorization after cache TTL expires (5 minutes) |
+
+## Requirements
+
+| Title | Description | Type |
+|---|---|---|
+| Envoy External Authorization API | authz_sidecar shall implement Envoy's External Authorization v3 gRPC API | Functional |
+| Role-based access control | Given user roles in `x-osmo-roles` header, authz_sidecar shall query PostgreSQL and validate access per role policies | Functional |
```
ecolternv, line 63: x-osmo-roles will be there when we use keycloak, but when we remove keycloak, this will need to read roles directly from postgres.
Will we also cache user->roles mapping? Or will this reintroduce a new database read for every request?

projects/auth/PROJ-148-direct-idp-integration.md
```diff
@@ -0,0 +1,1307 @@
...
+### Database Schema for User-Role Assignments
+
+```sql
+-- User-role assignment table
+CREATE TABLE user_roles (
+    id SERIAL PRIMARY KEY,
+    username VARCHAR(255) NOT NULL,
+    role_name VARCHAR(255) NOT NULL REFERENCES roles(name) ON DELETE CASCADE,
+    assigned_by VARCHAR(255) NOT NULL,
+    assigned_at TIMESTAMP WITH TIME ZONE DEFAULT NOW(),
```
ecolternv: Can this be `NOT NULL` as well?

projects/PROJ-148-auth-rework/PROJ-148-direct-idp-integration.md
```diff
@@ -0,0 +1,1307 @@
...
+osmo user roles add user@example.com --role osmo-ml-team
+
+# Assign a role with expiration
+osmo user roles add user@example.com --role osmo-ml-team --expires 2025-12-31
+
+# Remove a role from a user
+osmo user roles remove user@example.com --role osmo-ml-team
+
+# List all users with a role
+osmo role users list osmo-ml-team
```
ecolternv: Its kind of tricky that the first thing is singular and hte second thing is plural

`osmo role users`
`osmo user roles`

I can see myself getting confused and trying `osmo roles ...` and `osmo users ...`

projects/auth/PROJ-148-direct-idp-integration.md
```diff
@@ -0,0 +1,1307 @@
...
+    click.echo(f"Removed role '{role}' from {username}")
+```
+
+### PostgreSQL Functions
+
+Add the following functions to the PostgreSQL connector:
+
+**File**: `external/src/utils/connectors/postgres.py`
+
+```python
```
ecolternv: We really don't need this in the design document. This is more of the implementation, not the design. if there are some high level algorithms that are important to explain, then code/pseudocode makes sense, but I dont see why we need all of this here.

projects/PROJ-148-auth-rework/PROJ-148-resource-action-model.md
```diff
@@ -0,0 +1,870 @@
...
+│                                                                             │
+│  app                 Create, Read, Update,      (global)                    │
+│                      Delete, List                                           │
+│                                                                             │
+│  config              Read, Update               (global / service)          │
+│                                                                             │
+│  system              Health, Version            (public)                    │
+│                                                                             │
+│  internal            Operator, Logger, Router   (backend only)              │
+│                                                                             │
```
ecolternv, line 160: Does this allow us to make granular permissions such that:
- backend-operator can be allowed to connect and fetch jobs/push updates for only one single backend
- osmo-ctrl will have permission to push logs only for hte workflow that it belongs to?

projects/auth/PROJ-148-resource-action-model.md
```diff
@@ -0,0 +1,870 @@
...
+
+```
+<resource-type>/<identifier>
+
+Examples:
+  workflow/*           - All workflows
+  workflow/abc123      - Specific workflow
+  pool/default         - Default pool
+  pool/production/*    - Production pool and children
+  bucket/pool/default/* - Buckets in default pool
```
ecolternv: This doesn't match the pattern of 
```<resource-type>/<identifier>```
I would expect `bucket/my-bucket-name`. Why is bucket special?

File: projects/PROJ-148-auth-rework/PROJ-148-auth-sidecar.md [hunk 9]
```diff
@@ -0,0 +1,548 @@
...
+**Proposal Issue**: [#148](https://github.com/NVIDIA/OSMO/issues/148)
+
+## Overview
+
+The Authorization Sidecar project implements a high-performance Golang gRPC service that provides
+centralized role-based access control (RBAC) for all OSMO services through Envoy's External
+Authorization API. This replaces the Python `AccessControlMiddleware` with a more efficient,
+scalable, and maintainable solution.
+
+### Motivation
```
fernandol-nvidia, line 32: I think some of the motivation (out of context) might not fully explain your thought process. Perhaps it is worth adding a "Why Not?" section that very briefly outlines some consideration you've decided AGAINST... therefore giving light to the "Why" you are currently proposing.

projects/auth/PROJ-148-resource-action-model.md
RyaliNvidia: Oh. This was a leftover. Originally, I was playing around with the idea of buckets tied to a pool, but then removed it and this was leftover from when I was seeing how that looked

projects/PROJ-148-auth-rework/PROJ-148-resource-action-model.md
RyaliNvidia: Yes. I'll update the backend only field to be more aligned

projects/auth/PROJ-148-direct-idp-integration.md
RyaliNvidia: Sounds good. I had it there because it made it easier for me to fully see everything connect, but it is beyond the need a design doc

projects/PROJ-148-auth-rework/PROJ-148-direct-idp-integration.md
RyaliNvidia: Yeah. This section is marked as WIP as I just wanted to put some stuff down as an example, but this is not the final look. I'm still considering whether we even need like `osmo user` or just `osmo role`

projects/auth/PROJ-148-direct-idp-integration.md
RyaliNvidia: Yep. Updating now

projects/PROJ-148-auth-rework/PROJ-148-auth-sidecar.md
RyaliNvidia: When keycloak is removed, this will read roles from postgres **AND** `x-osmo-roles` header if the IDP supports roles. Once OSMO has user->role mapping, this will also be cached so it shouldn't introduce a new db read with every request

projects/auth/PROJ-148-auth-sidecar.md
RyaliNvidia: Yeah. I guess I wanted to mention that we keep the current sorta the same behavior. It doesn't belong in the Motivation section

projects/auth/PROJ-148-auth-sidecar.md
RyaliNvidia: I'll switch this to reference my performance numbers that I gather at the bottom. Those tests are from where both the middleware and the go code had caching and were operating on cache hits.
The main premise of the test is to compare the go code with gRPC against the current python which is on even playing ground

projects/auth/PROJ-148-direct-idp-integration.md
```diff
@@ -0,0 +1,1048 @@
...
+```
+
+---
+
+## Table of Contents
+
+1. [Prerequisites](#prerequisites)
+2. [Microsoft Entra ID (Azure AD)](#microsoft-entra-id-azure-ad)
+3. [Google OAuth2](#google-oauth2)
+4. [Amazon Cognito](#amazon-cognito)
```
vvnpn-nv: please add integration for okta this is one of the most commonly used SSO providers.

projects/PROJ-148-auth-rework/PROJ-148-auth-sidecar.md
RyaliNvidia: I do have a section below called `Alternatives Considered` which sorta does what you mentioned. Should I move this up to the top?

projects/auth/PROJ-148-auth-sidecar.md
RyaliNvidia: I think the biggest thing I need to figure out is how to connect the auth sidecar with the logging sidecar to send off logs.

But besides this, what are you concerns with using a sidecar in place of the middleware? From my perspective, the sidecar shouldn't actually add much overhead and actually provides more clarity to the service.

I do like the fact that we can just check the container to see the logs/health of the authorization logic. I didn't put this as a selling point because by default, we should add better logging to the logging sidecar for the authorization for no matter if we use the middleware or sidecar.

projects/auth/PROJ-148-direct-idp-integration.md
RyaliNvidia: Sounds good. I'll need to look into okta as I'm not too familiar with their auth process

vvnpn-nv: Instead of amazon cognito can we look into Identity Center (AWS SSO) instead

vvnpn-nv: another commonly used auth provider is [auth0](https://auth0.com/docs)

RyaliNvidia: Sounds good. I think I'll probably handle Azure AD, Google OAuth2, and AWS SSO for the first verification and then work on supporting the rest. I'll add a section detailing further auth providers we want to support

projects/PROJ-148-auth-rework/PROJ-148-auth-sidecar.md
fernandol-nvidia: Hmm nah, I think your latest motivation is actually good enough now :D 

Thank you.

projects/auth/PROJ-148-resource-action-model.md
```diff
@@ -0,0 +1,871 @@
...
+{
+  "name": "osmo-user",
+  "description": "Standard user role",
+  "policy": {
+    "statements": [
+      {
+        "effect": "Allow",
+        "actions": [
+          "workflow:*",
+          "task:Read", "task:Update", "task:Cancel", "task:Exec", "task:PortForward", "task:Rsync",
```
fernandol-nvidia: question, how do we define a policy for users only able to portforward tasks in their authorized pool?

projects/PROJ-148-auth-rework/PROJ-148-direct-idp-integration.md
```diff
@@ -0,0 +1,1048 @@
...
+   - Troubleshooting authentication flows across three systems (IDP → Keycloak → OSMO)
+
+5. **User Experience Friction** — The extra hop through Keycloak adds:
+   - Additional redirect latency during login
+   - Potential for session mismatches between Keycloak and OSMO
+   - Confusing logout behavior (must log out of both systems)
+
+**By removing Keycloak and connecting Envoy directly to the IDP**, we achieve:
+
+- **Single source of truth** for role assignments (OSMO database only)
```
fernandol-nvidia, line 75: question, if an IDP also provides roles, do we want to support that? Or is everything done strictly thru OSMO?

projects/PROJ-148-auth-rework/PROJ-148-auth-sidecar.md
```diff
@@ -0,0 +1,551 @@
...
+FROM roles
+WHERE name = ANY($1)
+ORDER BY name;
+```
+
+### 3. Role Caching
+
+**Implementation**: Thread-safe in-memory cache with LRU eviction
+
+**Cache Key**: Sorted, comma-separated role names (e.g., `"osmo-default,osmo-user"`)
```
fernandol-nvidia, line 174: can you share what the cache value might look like?

projects/auth/PROJ-148-resource-action-model.md
RyaliNvidia: Ah. So task also uses the pool resource to determine the authorized pool as per the section above. Let me try to fix some examples or update the wording

projects/PROJ-148-auth-rework/PROJ-148-auth-sidecar.md
RyaliNvidia: In my test for this, the cache values look like: https://github.com/NVIDIA/OSMO/pull/126/files#diff-0ec6cf70f862a1ea33cc5b4a3973ccca4e83f3dee324794ba2c327c8ccbd0897R44

projects/PROJ-148-auth-rework/PROJ-148-direct-idp-integration.md
RyaliNvidia: Yes the IDP provides roles and we will use that. Ideally to me, we would fetch the IDP roles and then update our DB with the information, so our DB is entirely accurate. As a result, the DB is still the single source of truth for OSMO. However, this option has an issue where if the user has been removed from access in the IDP, OSMO doesn't have real-time trigger to be aware of this so the OSMO DB will only be updated when the user tries to login next.

The other option is from @ecolternv 's suggestion that we don't store the role provided by the IDP and when auth is evaluated, we used the combine IDP value + OSMO specified config to determine access. This removes OSMO's responsibility to be up to date, but I don't really like that we can't tell the user's current access ability. Either way is a simple change, I prefer the first option, but I'm open to either

projects/auth/PROJ-148-resource-action-model.md
RyaliNvidia: Updated

projects/auth/PROJ-148-direct-idp-integration.md
RyaliNvidia: Updated the content

projects/PROJ-148-auth-rework/PROJ-148-resource-action-model.md
```diff
@@ -0,0 +1,837 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# Resource-Action Permission Model Design
```
vvnpn-nv, line 19: can we leverage existing libraries such as Casbin for RBAC control 
https://casbin.org

RyaliNvidia: I forgot that I actually looked into this as @fernandol-nvidia referenced this to me a couple of weeks ago. The reason why I strayed away from it is because I thought it had a strict compile time definition of the role to policy mapping, but I think I misunderstood. I'll play around with Casbin and see if we can leverage this

projects/auth/PROJ-148-auth-sidecar.md
RyaliNvidia: Diving in more, I believe that middleware could be where the auth logic lies in the future. However, with the current version of OSMO, a sidecar allows us not to worry about the Go vs Python implementation. Once we are just using Go, then the middleware transition should be easy as it is just bringing in the logic into the service.

projects/PROJ-148-auth-rework/PROJ-148-resource-action-model.md
RyaliNvidia: I did a deep dive into Casbin and wrote a demo service using Casbin. The conclusion from it was that Casbin is good if we had nothing and needed to start on something, but using Casbin at this point doesn't simplify any of the work we would need to do with auth. We would need to write the same amount of logic. It would simply just be a syntax language like OPA we would write the logic in, but the complexity of going into the db to fetch information will still need to be written by us either way.

Casbin would just be another 3rd party library to rely on

PR #167, opened 2025-12-18
Efficient Workflow Cleanup through Using Async Operations for Log Migration

ethany-nv:
## Description
This PR uses asyncio to migrate logs for all the tasks within a workflow concurrently. Previously, the worker would synchronously migrate logs from Redis to cloud storage per task, which was inefficient.

This PR also batches together delete operations to increase the speed of clearing all the stale Redis keys.

Issue #174 

### Experiment

To see the performance gain of this optimization, we measure the duration of the `CleanupWorkflow` task when running large group workflows on two different OSMO instances - one without and one with the optimization. For each workflow ran, each task outputs 5000 lines.

| Num Tasks | W/O Optimization | With Optimization |
|----------:|------------------:|------------------:|
|        50 |              3:40 |              1:46 |
|       100 |              5:44 |              3:21 |
|       200 |    N/A (Probe Kill) |            5:58 |

## Checklist


src/utils/job/jobs.py
```diff
@@ -1320,121 +1323,114 @@ def execute(self, context: JobExecutionContext,
...
+        async def run_log_migrations():
+            await asyncio.gather(
+                *(
+                    migrate_logs_concurrently(redis_url, redis_key, file_name)
+                    for redis_url, redis_key, file_name in task_parameters
+                )
+            )
 
-                    migrate_logs(workflow_obj.logs, f'{prefix}-error-logs', task_error_log_name)
+        asyncio.run(run_log_migrations())
```
xutongNV, line 1447: are they concurrent? should it be like asyncio.create_task(...) and then asyncio.gather(...)

src/utils/job/jobs.py
```diff
@@ -1320,121 +1323,114 @@ def execute(self, context: JobExecutionContext,
...
+                    redis_url,
+                    redis_key,
+                    temp_file.name,
                 )
 
-                last_timestamp = update_progress_writer(
-                    progress_writer,
-                    last_timestamp,
-                    progress_iter_freq)
+                await progress_writer.report_progress_async()
```
xutongNV, line 1391: can you add more progress write below as well

src/utils/job/jobs.py
ethany-nv: `run_log_migrations()` has `asyncio.gather()`. When I pass all the coroutines to `gather()`, it implicitly turns each lambda into an asyncio task. We only need to use `asyncio.create_task` if we want the `task` handle that is returned from that function.

xutongNV (approved):
have you test this with a big workflow?

RyaliNvidia (dismissed):

src/utils/job/jobs.py
```diff
@@ -1320,121 +1323,124 @@ def execute(self, context: JobExecutionContext,
...
+                await asyncio.to_thread(_upload_logs)
 
-                task_redis_path = common.get_redis_task_log_name(
-                    self.workflow_id,
-                    task_obj.name,
-                    task_obj.retry_id,
-                )
+                await progress_writer.report_progress_async()
+
+        semaphore = asyncio.Semaphore(10)
```
fernandol-nvidia: let's not hardcode and please tie this to the storage_client executor options so we are aligned

src/utils/job/jobs.py
```diff
@@ -1305,6 +1306,8 @@ def execute(self, context: JobExecutionContext,
 
         redis_client = redis.from_url(workflow_obj.logs)
 
+        redis_batch_pipeline = redis_client.pipeline()
```
fernandol-nvidia, line 1311: FYI, there is an async flavor of this as well. But I am not sure if we are ready for it (i.e. is LogStreamBody capable of async). Something to consider...

ethany-nv: Ok, noted

PR #177, opened 2025-12-19
Perform Client Side Data Auth Check In the Event of Environment Based Auth

fernandol-nvidia:
Issue #153

## Checklist

### DIFF
--- a/src/cli/dataset.py
+++ b/src/cli/dataset.py
@@ -24,7 +24,6 @@ import hashlib
 import logging
 import os
 import re
-import sys
 import textwrap
 from typing import Any, Dict, List
 import yaml


src/cli/dataset.py
```diff
@@ -954,6 +957,73 @@ def _run_migrate_command(service_client: client.ServiceClient, args: argparse.Na
...
+                    access_type=storage_lib.AccessType.READ,
+                )
+            case _:
+                storage_backend.data_auth(
+                    data_cred=data_cred,
+                )
+
+        # Auth check passed
+        print(json.dumps({'status': 'pass'}))
+        sys.exit(0)
```
RyaliNvidia: NIT: Seems like the sys.exit is unnecessary? for this and the credential error

src/cli/data.py
```diff
@@ -179,6 +182,67 @@ def _run_delete_command(service_client: client.ServiceClient, args: argparse.Nam
...
+        sys.exit(0)
+
+    except osmo_errors.OSMOError as err:
+        # Execution error (service issue, network problem, etc.)
+        print(json.dumps({'status': 'error', 'error': str(err)}))
+        sys.exit(1)
+
+    except Exception as err:  # pylint: disable=broad-except
+        print(json.dumps({'status': 'error', 'error': f'Unexpected error: {str(err)}'}))
+        sys.exit(1)
```
RyaliNvidia: This is more of a design discussion with the CLI, but these errors (OSMOError and general Exception) is normally caught by the cli.py for it to determine the correct sys.exit. This convolutes the behavior a bit

src/cli/dataset.py
fernandol-nvidia: Yup, unnecessary. Removed

src/cli/data.py
fernandol-nvidia: Good catch, I think we can simply remove this and just reuse the CLI error handling. Any errors will have exit code != 0 anyways.

PR #208, opened 2026-01-06
Operator Redesin POC

xutongNV:
## Description

Issue #147 

## Checklist

### DIFF
--- a/src/service/agent/agent_worker.py
+++ b/src/service/agent/agent_worker.py
@@ -16,7 +16,6 @@ limitations under the License.
 SPDX-License-Identifier: Apache-2.0
 """
 
-import datetime
 import json
 import logging
 import os
@@ -114,8 +113,12 @@ class AgentWorker:
 
             # Process the message based on type
             if message.type == backend_messages.MessageType.UPDATE_POD:
-                helpers.queue_update_group_job(self.postgres,
-                                               backend_messages.UpdatePodBody(**message.body))
+                if isinstance(message.body, dict):
+                    helpers.queue_update_group_job(self.postgres,
+                                                   backend_messages.UpdatePodBody(**message.body))
+                else:
+                    logging.error('Invalid body type for UPDATE_POD message: %s',
+                                  type(message.body))
             else:
                 logging.error('Ignoring invalid backend listener message type %s, uuid %s',
                               message.type.value, message.uuid)
@@ -221,7 +224,7 @@ class AgentWorker:
                     continue
 
                 # Process each message
-                for stream_name, stream_messages in messages:
+                for _, stream_messages in messages:
                     for message_id, message_data in stream_messages:
                         if b'message' in message_data:
                             message_json = message_data[b'message'].decode('utf-8')
--- a/src/service/agent/helpers.py
+++ b/src/service/agent/helpers.py
@@ -1,5 +1,6 @@
 """
-SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES.
+All rights reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -20,7 +21,6 @@ import asyncio
 import datetime
 import json
 import logging
-import time
 from typing import Dict
 from urllib.parse import urlparse
 
--- a/src/utils/backend_messages.py
+++ b/src/utils/backend_messages.py
@@ -1,5 +1,6 @@
 """
-SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES.
+All rights reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/src/utils/metrics/metrics.py
+++ b/src/utils/metrics/metrics.py
@@ -1,5 +1,6 @@
 """
-SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES.
+All rights reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.


src/operator/utils/helpers.go
```diff
@@ -0,0 +1,147 @@
...
+// UnackMessages provides flow control and reliability by tracking messages that have been sent
+// but not yet acknowledged by the server. When the max limit is reached, sending is
+// paused until ACKs are received, preventing unbounded memory growth.
+//
+// Key features:
+// - Thread-safe message tracking using mutex
+// - Flow control via readyToSend channel
+// - Automatic backpressure when max unacked messages limit is reached
+// - Message persistence for potential reconnection scenarios
+type UnackMessages struct {
```
ecolternv: Can we see if there is a way to avoid having `helpers.go` files but rather to have different files for each major self-contained functionality/struct?

Maybe we can keep helpers.go if there is no where else to put `ParseServiceURL`, but it seems like most of this could be in an `unacked_messages.go` file.

src/operator/workflow_listener.go
```diff
@@ -0,0 +1,559 @@
...
+		// Clean exit
+		break
+	}
+
+	log.Println("Workflow Listener stopped gracefully")
+}
+
+// runWorkflowListener establishes a connection to the operator service
+// and maintains a bidirectional stream
+func runWorkflowListener(
```
ecolternv: Does it make sense to make WorkflowListener into a struct with this and receiveMessages and sendMessages etc. as methods on that struct? It seems like there are a lot of other functions that this one calls that all need the same parameters, like ctx context.Context, utils.UnackMessages, etc, And it the list will probably grow as we add more messages to this.

src/operator/workflow_listener.go
```diff
@@ -0,0 +1,559 @@
...
+		return nil, fmt.Errorf("failed to marshal pod update: %w", err)
+	}
+
+	// Generate random UUID (matching Python's uuid.uuid4().hex format)
+	messageUUID := strings.ReplaceAll(uuid.New().String(), "-", "")
+
+	msg := &pb.ListenerMessage{
+		Type:      pb.ListenerMessage_update_pod,
+		Uuid:      messageUUID,
+		Timestamp: time.Now().UTC().Format("2006-01-02T15:04:05.999999"),
```
ecolternv, line 525: Is this just a placeholder?

src/operator/workflow_listener_test.go
```diff
@@ -0,0 +1,1187 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION. All rights reserved.
```
ecolternv: 2026! I think some other copyrights are wrong too

src/operator/workflow_listener_test.go
```diff
@@ -0,0 +1,1187 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			result := utils.CalculatePodStatus(tt.pod).Status
+			if result != tt.expected {
+				t.Errorf("CalculatePodStatus() = %v, expected %v", result, tt.expected)
+			}
+		})
+	}
+}
+
+// Test getPodMessage function - COMMENTED OUT: getPodMessage is now internal to utils.CalculatePodStatus
```
ecolternv: Can we remove this commented out test, or move it to somewhere in utils since thats now where CalculatePodStatus is?

src/proto/operator/messages.pb.go
```diff
@@ -0,0 +1,431 @@
+// Code generated by protoc-gen-go. DO NOT EDIT.
```
ecolternv: If this code is generated then we should probably avoid checking it in. Can we have the protobuffer generation step run as part of the bazel build process, and the outputs go in the bazel outputs? We should still be able to use generated files like that in our go application

src/proto/operator/messages.proto
```diff
@@ -0,0 +1,65 @@
...
+//SPDX-License-Identifier: Apache-2.0
+
+syntax = "proto3";
+
+package operator;
+
+option go_package = "go.corp.nvidia.com/osmo/proto/operator;operator";
+
+// ListenerMessage represents a message exchanged between the operator service and workflow backends
+message ListenerMessage {
```
ecolternv, line 24: We should use separate type for messages from Server -> Client and messages from Client -> Server

src/proto/operator/messages.proto
```diff
@@ -0,0 +1,65 @@
...
+  // Possible message types for listener communication
+  enum MessageType {
+    unspecified = 0;
+    ack = 1;           // Acknowledgment response from operator to backend
+    update_pod = 2;    // Pod update from backend to operator
+  }
+
+  MessageType type = 1;
```
ecolternv, line 31: Instead of MessageType and `body`,  you can use protobuf's `oneof`.

src/service/agent/agent_worker.py
```diff
@@ -0,0 +1,266 @@
+"""
```
ecolternv, line 1: Is this separate from `worker.py` because golang doesn't support kombu, so the agenet_worker needs to pull jobs off of redis in a different way?

src/operator/workflow_listener.go
xutongNV: This is a format string, in GO you cant do "YYYY-MM-DDxxxxx". The actual numbers (2006, 01, 02, etc.) are hardcoded constants in the Go language design. They represent the reference date: January 2, 2006 at 3:04:05 PM

src/service/agent/agent_worker.py
xutongNV: Yes

src/proto/operator/messages.pb.go
xutongNV: The go code already depends on the proto lib and auto generates the deps. I accidentally checked this in

src/proto/operator/messages.proto
xutongNV: currently we have to keep type and body fields to align with python side
```
class MessageBody(pydantic.BaseModel, extra=pydantic.Extra.forbid):
    """
    Used for Message Structure
    """
    type: MessageType
    body: Dict | LoggingBody
    uuid: str = pydantic.Field(default_factory=common.generate_unique_id)
    timestamp: datetime.datetime = pydantic.Field(default_factory=common.current_time)
 ```

File: src/operator/workflow_listener.go (hunk 2 above)
Comment (xutongNV) [reply src/operator/workflow_listener.go]: greate idea. we can also have resourceListener etc in the future

File: src/proto/operator/messages.proto (hunk 8 above)
Comment (elookpotts-nvidia) [reply src/proto/operator/messages.proto:31]: Can we have a transform on the python side so that we can use oneof? It will be quite challenging to switch later on, and `oneof` is the protobuf way to do this.

File: deployments/brev/README.md [hunk 10]
```diff
@@ -18,7 +18,8 @@ SPDX-License-Identifier: Apache-2.0
 
 # OSMO Brev Deployment
 
-[![ Click here to deploy.](https://brev-assets.s3.us-west-1.amazonaws.com/nv-lb-dark.svg)](https://brev.nvidia.com/launchable/deploy?launchableID=env-36a6a7qnkOMOP2vgiBRaw2e3jpW)
+[![NVIDIA-OSMO](https://img.shields.io/badge/NVIDIA-OSMO-76b900?logo=nvidia)](https://github.com/NVIDIA/OSMO)
+[![Deploy on Brev](https://brev-assets.s3.us-west-1.amazonaws.com/nv-lb-dark.svg)](https://brev.nvidia.com/launchable/deploy?launchableID=env-36a6a7qnkOMOP2vgiBRaw2e3jpW)
```
elookpotts-nvidia, line 22: Looks like `xutongr/code` was branched off of main and is bringing in commits from main - can you branch off of the feature branch instead to keep this PR focused?

xutongNV: I rebased on main to make feature branch up to date. Do you want a separate PR to do this?

elookpotts-nvidia: Gotcha, no worries for this PR, but typically a separate PR will make this easier (1) for reviewing changes and (2) if there are any issues with the rebase and we need to do branch surgery. We merge with squash commits so things can get messy quickly if a rebase goes wrong.

src/proto/operator/messages.proto
xutongNV: I can have this in future PRs

PR #209, opened 2026-01-08
Connection test to service should retry on 429 and 503

tdewanNvidia:
## Description
Connection test to service should retry on 429 and 503
Issue - None

## Checklist

### DIFF
--- a/src/operator/utils/node_validation_test/connection_validator.py
+++ b/src/operator/utils/node_validation_test/connection_validator.py
@@ -37,10 +37,15 @@ class URLTestConfig(pydantic.BaseModel):
     method: str = pydantic.Field(default='GET', description='HTTP method to use')
     timeout: int = pydantic.Field(
         default=30, description='Timeout in seconds for the connection test')
-    expected_status_code: int = pydantic.Field(
-        default=200, description='Expected HTTP status code')
+    expected_status_code: Optional[int] = pydantic.Field(
+        default=None, description='Expected HTTP status code (None means any non-5xx is success)')
     condition_name: Optional[str] = pydantic.Field(
         default='ServiceConnectionTestFailure', description='Custom condition name for this URL')
+    retriable_status_codes: List[int] = pydantic.Field(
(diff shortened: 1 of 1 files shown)


src/operator/utils/node_validation_test/connection_validator.py
```diff
@@ -128,21 +140,50 @@ def _connection_test(self, url_config: URLTestConfig) -> test_base.NodeCondition
...
+                    return None
+            else:
+                # Check if status code is retriable (e.g., 429 rate limiting, 503 unavailable)
+                if status_code in url_config.retriable_status_codes:
+                    logging.warning(
+                        'Retriable status code from %s: %s, will retry',
+                        url_config.url,
+                        status_code,
+                    )
+                    return None
```
RyaliNvidia, line 163: How is the behavior different between this and failure_status_codes since they both return None. How do we differentiate between the two to retry with this one but not with the one below?

src/operator/utils/node_validation_test/connection_validator.py
```diff
@@ -37,10 +37,15 @@ class URLTestConfig(pydantic.BaseModel):
...
-    expected_status_code: int = pydantic.Field(
-        default=200, description='Expected HTTP status code')
+    expected_status_code: Optional[int] = pydantic.Field(
+        default=None, description='Expected HTTP status code (None means any non-5xx is success)')
     condition_name: Optional[str] = pydantic.Field(
         default='ServiceConnectionTestFailure', description='Custom condition name for this URL')
+    retriable_status_codes: List[int] = pydantic.Field(
+        default=[429, 503], description='Status codes that should trigger retry')
+    failure_status_codes: List[int] = pydantic.Field(
+        default=[500, 501, 502, 504, 505, 506, 507, 508, 510, 511],
```
ecolternv, line 47: Seems kind of odd that by default this is full of 500 status codes, but down below we have
```
 if status_code in url_config.failure_status_codes or status_code >= 500:
```
This might as well default to an empty list if we always check all 500 status codes.

Issue #206, opened 2026-01-06
Project: Native Support for Multi-node NvLink


ecolternv:
### Have you read the Project Process docs?

- [x] Yes, I have read and understood the RFC docs

### Summary

- Add support for creating [ComputeDomain CRDs](https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/dra-cds.html#computedomains-multi-node-nvlink-simplified) natively for workflows that have NVL72 or other multi-node nvlink enabled backends.
- Add support for topology aware scheduling introduced in KAI Scheduler v0.10 to allow users to scheduled all of their tasks of a given workflow on the same NVL72 rack.

### Person in Charge (PIC)

@ecolternv

### Motivation

The current Blackwell generation of GPUs with GB200 and GB300, and future announced GPU generations feature Multi Node NVLINK with NVL72 and even higher numbers like NVL144, etc.

To properly get full node to node performance for multi-node training, workloads scheduled in kubernetes need to take advantage of NvLink, and carefully control how they nodes are placed in the cluster so they end up in the same racks.

### Problem

**ComputeDomain CRD:**
To use MultiNode NvLink in kubernetes, you must create a ComputeDomain CRD and all pods that are part of a given training run must have a resourceClaim that points to the same ComputeDomain. OSMO does not currently support creating/destroying ComputeDomains along with the lifecycle of the workflow.

**Topology Aware Scheduling:**
To use MultiNode NvLink, pods that wish to communicate using NvLink must be placed in the same NVL72/NVL144 rack by the scheduler.

For very large training runs, users may want full control over the placement of tasks in racks, for example a training run may be 4 x tensor parallel, 4 x pipeline parallel, and 4 x data parallel. This means there will be 64 pods total: 4 groups of 16 that each represent a single instance of a model. A user may wish to ensure that each group of 16 pods that represent an instance of the model all end up in the same NVL72 rack.

Currently OSMO has some ability to handle topology aware scheduling with podAffinities added to OSMO pod templates, but this doesn't provide the granular level of control needed for many usecases in an NVLink enabled cluster.

### References

- Compute Domain CRD: https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/dra-cds.html
- KAI Scheduler topology aware scheduling docs: https://github.com/NVIDIA/KAI-Scheduler/tree/main/docs/topology

PR #211, opened 2026-01-08
Add new project proposal to describe nvlink + topology aware scheduling

ecolternv:
## Description
Add project design document for nvlink + topology aware scheduling support

Issue #206 

## Checklist

### DIFF
--- a/projects/PROJ-206-nvlink-support/PROJ-206-topology-aware-scheduling.md
+++ b/projects/PROJ-206-nvlink-support/PROJ-206-topology-aware-scheduling.md
@@ -845,4 +845,4 @@ the PodGroup and other CRDs are created and interpreted correctly by KAI schedul
 
 ## Open Questions
 
-- [ ] TBD
+- [ ] How can we visualize the racks/spines/zones in the UI to help users better understand why workflows aren't scheduling.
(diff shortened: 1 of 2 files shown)


RyaliNvidia:
This is a more general question regarding the feature, but once the feature is there, how will you verify that NVLink is performing as expected? What kind of tests can you do to verify that NVLink is actually being used and whether it is performing as well as expected? It could be that there isn't an issue with the implementation and that NVLink is not performing well, but it would be nice to know that.

RyaliNvidia:
Is NVLink a feature we want users to opt into so that even if the feature is there, they can specify not to use it? This is because they might want to build confidence that NVLink is improving their performance so they would want to run to side by side tests with NVLink there and not. 

This way, we could get a KPI on how many people are switching to using NVLink and the usefulness of this feature once deployed

projects/PROJ-206-nvlink-support/PROJ-206-topology-aware-scheduling.md
```diff
@@ -0,0 +1,846 @@
...
+
+
+```yaml
+{
+    "name": "my-pool-01",
+    "description": "A pool with topology aware scheduling enabled",
+    # Topology keys that appear first are the finest grain
+    # (Ie multiple racks belong in the same spine)
+    "topology_keys": [
+        "topology.kubernetes.io/rack",
```
RyaliNvidia: Consider making this a key value pair where the key is the user facing field and the value is the value in kubernetes like `"topology.kubernetes.io/rack"`

RyaliNvidia:
Another thing to consider is if we can change the resources list to show the zones/racks/whatnot visually so users know why their workflow isn't scheduling even though there are resources available.

projects/PROJ-206-nvlink-support/PROJ-206-topology-aware-scheduling.md
ecolternv: Done

PR #216, opened 2026-01-12
update proto to use oneof

xutongNV:
## Description

Issue #147

## Checklist

### DIFF
--- a/src/proto/operator/messages.proto
+++ b/src/proto/operator/messages.proto
@@ -20,23 +20,34 @@ package operator;
 
 option go_package = "go.corp.nvidia.com/osmo/proto/operator;operator";
 
+// LoggingType represents different logging levels
+enum LoggingType {
+  INFO = 0;
+  DEBUG = 1;
+  WARNING = 2;
+  EXCEPTION = 3;
+}
+
+// LoggingBody represents a logging message
+message LoggingBody {
+  LoggingType type = 1;
+  string text = 2;
+  string workflow_uuid = 3;  // Empty string represents non-workflow logging
+}
+
 // ListenerMessage represents messages sent from workflow backends to the operator service
 message ListenerMessage {
-  // Possible message types for listener communication
-  enum MessageType {
-    unspecified = 0;
-    update_pod = 1;    // Pod update from backend to operator
-  }
-
-  MessageType type = 1;
-
   // UUID uniquely identifies this message for correlation and acknowledgment
-  string uuid = 2;
+  string uuid = 1;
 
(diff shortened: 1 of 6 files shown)


src/proto/operator/messages.proto
```diff
@@ -20,23 +20,34 @@ package operator;
...
   // Timestamp when the message was created (ISO 8601 format, e.g., "2025-01-15T10:30:00Z")
-  string timestamp = 3;
+  string timestamp = 2;
 
-  string body = 4;
+  // Message body - exactly one of these must be set
+  oneof body {
+    UpdatePodBody update_pod = 3;
+    LoggingBody logging = 4;
+  }
```
elookpotts-nvidia, line 50: [nit] I tend to prefer nested messages in protobufs instead of a flat file for readability - then you don't need to rely on the `Body` suffix to indicate what the messages are, and consumers understand the scoping of the message.

src/proto/operator/messages.proto
```diff
@@ -20,23 +20,34 @@ package operator;
...
-  string uuid = 2;
+  string uuid = 1;
 
   // Timestamp when the message was created (ISO 8601 format, e.g., "2025-01-15T10:30:00Z")
-  string timestamp = 3;
+  string timestamp = 2;
 
-  string body = 4;
+  // Message body - exactly one of these must be set
+  oneof body {
```
elookpotts-nvidia, line 47: [nit] I personally prefer `payload`, since this isn't equivalent to a HTTP request body

elookpotts-nvidia (approved):

src/proto/operator/messages.proto
elookpotts-nvidia: There is a performance tradeoff - the sub-messages can introduce overhead due to the increased message size.

src/proto/operator/messages.proto
```diff
@@ -20,23 +20,34 @@ package operator;
...
 
+// ListenerMessage represents messages sent from workflow backends to the operator service
+message ListenerMessage {
   // UUID uniquely identifies this message for correlation and acknowledgment
-  string uuid = 2;
+  string uuid = 1;
 
   // Timestamp when the message was created (ISO 8601 format, e.g., "2025-01-15T10:30:00Z")
-  string timestamp = 3;
+  string timestamp = 2;
```
fernandol-nvidia, line 44: Do we want to be strict about this with: https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/timestamp.proto

?

Issue #220, opened 2026-01-14
Project: Dynamic Pool Quota Managment


fernandol-nvidia:
### Have you read the Project Process docs?

- [x] Yes, I have read and understood the RFC docs

### Summary

Allow pool admins to be able to configure granular sub-pool quotas to ensure fair share.

### Person in Charge (PIC)

@fernandol-nvidia

### Motivation

Gives a large team the ability to partition team quota as it sees fit (e.g. person, project, sub-teams, etc)

### Problem

Currently, a single pool has a single over-arching quota (and a single queue). If everyone on the team submits HIGH priority with a deep backlog of workflows, then we run the risk of disproportional resource consumption.

A pool admin should be able to partition the pool quota by business objectives (e.g. sub teams, projects, sub pools, etc). This allows a sub-pool to have deep backlog of workflows queued up without worrying about unfair resource consumption across the entire pool.

PR #221, opened 2026-01-14
#220 - Design documentation for dynamic subpool

fernandol-nvidia:
## Description
Design documentation for dynamic subpool

Issue #220 

## Checklist

### DIFF
--- a/projects/PROJ-220-dynamic-subpool.md
+++ b/projects/PROJ-220-dynamic-subpool.md
@@ -214,6 +214,8 @@ osmo workflow submit --pool <parent--subpool> my-workflow.yaml
 
 Example `osmo pool list` output (HIGH/NORMAL only; Available can be negative while draining):
 
+> **Note**: Only subpools in states `ACTIVE` and `DELETING` are shown in this table. `ARCHIVED` subpools are omitted to ensure GPU quota accounting remains balanced and clear.
+
 ```
 Pool            Status         Subpool State  GPU Quota        Used  Available
 --------------  -------------  -------------  ---------------  ----  ---------
@@ -222,6 +224,7 @@ team            ONLINE         -              10 (Total: 100)  50    -40
 ├─ team--b      ONLINE         ACTIVE         40               10    30
 └─ team--c      ONLINE         ACTIVE         20               0     20
 ```
+
 
 Column semantics:
 - **GPU Quota (parent)**: `unallocated (Total: parent_quota)`
@@ -268,10 +271,18 @@ At all times: `_shared.quota + sum(active_subpool_quotas) = parent.quota`
 ### Subpool State
 
 ```
-┌──────────┐            ┌────────┐   delete   ┌───────────┐   drained  ┌──────────┐
-│  CREATE  │ ─────────► │ ACTIVE │ ─────────► │ DELETING  │ ─────────► │ ARCHIVED │
-└──────────┘            └────────┘            └───────────┘            └──────────┘
+  create    ┌────────┐   delete   ┌───────────┐   drained  ┌──────────┐
+  ────────► │ ACTIVE │ ─────────► │ DELETING  │ ─────────► │ ARCHIVED │
+            └────────┘            └───────────┘            └──────────┘
+                ▲                                                  │
+                │                                                  │
+                └──────────────── recreate ────────────────────────┘
 ```
+
+**State behavior**:
+- **ACTIVE**: All operations allowed (read, update quota, delete, submit jobs)
+- **DELETING**: Read-only. No updates allowed (cannot delete, update quota, or recreate) until drained to ARCHIVED
+- **ARCHIVED**: Read-only. Can only recreate to transition back to ACTIVE
 
 ### Inheritance (subpools are quota partitions only)
 
@@ -386,21 +397,22 @@ ALTER TABLE pools ADD COLUMN subpool_state TEXT;                        -- NULL 
 
 ### API Endpoints
 
-Subpools are pool entries in the `pools` table, so **read operations** use existing pool endpoints
-with the canonical name (`{parent}--{subpool}`). **Write operations** use dedicated `/subpool`
-endpoints to enforce parent context, quota constraints, and state-machine semantics.
+Subpools are pool entries in the `pools` table, but **all operations** use dedicated `/subpool`
+endpoints to enforce parent context, maintain consistent API semantics, and clearly distinguish
+subpool-specific behavior (quota constraints, state-machine semantics).
 
 | Endpoint | Method | Description |
 |----------|--------|-------------|
-| `/api/configs/pool/{parent}--{subpool}` | GET | Read subpool (existing pool endpoint) |
+| `/api/configs/pool/{parent}/subpool/{subpool}` | GET | Read subpool |
 | `/api/configs/pool/{parent}/subpool` | GET | List subpools of parent |
 | `/api/configs/pool/{parent}/subpool` | POST | Create subpool under parent |
 | `/api/configs/pool/{parent}/subpool/{subpool}` | PATCH | Update quota (quota-only) |
 | `/api/configs/pool/{parent}/subpool/{subpool}` | DELETE | Delete (smart, state-managed) |
 
-**Design rationale**: Write operations have different semantics than top-level pools (quota-only
-updates, parent-constrained creates, state-machine deletes), so they use dedicated endpoints.
-Read operations work with the canonical name since subpools are stored as pool entries.
+**Design rationale**: All operations use dedicated `/subpool` endpoints to provide consistent
+API semantics, enforce parent context, clearly distinguish subpool-specific behavior
+(quota-only updates, parent-constrained creates, state-machine deletes) from top-level pool operations,
+and simplify authorization enforcement by making pool and subpool access controls distinct and explicit.
 
 ### Security
 
@@ -487,7 +499,7 @@ KAI queue name, which is not visible to users.
 
 #### Migration Path
 
-No migration is required for existing pools:
+No migration is required for existing workflows:
 1. On deployment, OSMO will create `--_shared` queues for all existing pools
 2. Existing workflows continue running in their original queues until completion
 3. New submissions route to `--_shared` queues automatically


projects/PROJ-220-dynamic-subpool.md
```diff
@@ -0,0 +1,944 @@
...
+Solution: Create an explicit leaf queue ("shadow queue") for parent submissions
+
+Before (problematic):
+  osmo-default → osmo-pool-team (parent AND leaf, quota=100)
+                 └── Jobs submitted here can consume child quota
+
+After (shadow queue pattern):
+  osmo-default → osmo-pool-team (parent container, quota=100)
+                 ├── osmo-pool-team--_shared (leaf, quota=10)  ← parent submissions go here
+                 ├── osmo-pool-team--a (leaf, quota=30)
```
RyaliNvidia, line 152: I'm not sure how I feel about the `--` delimiter right now to differentiate pool and subpool. I feel like it is a little inelegant, but something we can think about later

projects/PROJ-220-dynamic-subpool.md
```diff
@@ -0,0 +1,944 @@
...
+osmo-default-{namespace}
+└── osmo-pool-{namespace}-team (parent container, quota=100)
+    └── osmo-pool-{namespace}-team--_shared (leaf, quota=100)
+```
+
+**Pool WITH subpools** (shadow queue gets unallocated portion):
+
+```
+osmo-default-{namespace}
+└── osmo-pool-{namespace}-team (parent container, quota=100)
```
RyaliNvidia, line 308: Right now, quotas do not need to be hard defined. For example, the quota can be left empty which means that as nodes come in and out of the cluster, they can be used for NORMAL/HIGH priority jobs.

How will subpools and _shared pools cover this? Can it cover this case?

projects/PROJ-220-dynamic-subpool.md
```diff
@@ -0,0 +1,944 @@
...
+
+### API Endpoints
+
+Subpools are pool entries in the `pools` table, so **read operations** use existing pool endpoints
+with the canonical name (`{parent}--{subpool}`). **Write operations** use dedicated `/subpool`
+endpoints to enforce parent context, quota constraints, and state-machine semantics.
+
+| Endpoint | Method | Description |
+|----------|--------|-------------|
+| `/api/configs/pool/{parent}--{subpool}` | GET | Read subpool (existing pool endpoint) |
```
RyaliNvidia: I feel like this should be `/api/configs/pool/{parent}/subpool/{subpool}` GET

projects/PROJ-220-dynamic-subpool.md
```diff
@@ -0,0 +1,944 @@
...
+  patterns (e.g. `osmo-team-*`) to naturally cover subpools via prefix matching.
+
+### Design Decisions
+
+| Decision | Choice | Rationale |
+|----------|--------|-----------|
+| Shadow queue for all pools? | **Yes (`--_shared`)** | KAI requires leaf queues for job submissions; consistent pattern for all pools regardless of whether they have subpools |
+| Shadow queue visibility? | **Hidden** | Users submit to `team`, OSMO routes to `team--_shared` internally; simplifies UX |
+| Reserved subpool names? | **`_shared`** | Prevent user collision with internal shadow queue; names starting with `_` are reserved |
+| Inherit all config from parent? | **Yes** | No drift, one source of truth |
```
RyaliNvidia, line 439: This makes it easier for us to implement, but have you confirmed that this is expected for users and that they won't expect it to be different per subpool

projects/PROJ-220-dynamic-subpool.md
```diff
@@ -0,0 +1,944 @@
...
+| Subpool-specific user access? | **No** | Simpler, inherits from parent |
+| Retention policy for ARCHIVED? | **Never hard delete** | Audit trail; no garbage collection |
+| Notifications on state transitions? | **No** | Status visible in CLI |
+| Subpool names globally unique? | **No** (scoped) | Avoid cross-team name collisions; use canonical `{parent}--{subpool}` for uniqueness |
+| Rounding | **floor** | Whole GPU quotas only; remainder stays in parent unallocated |
+| Delimiter in pool names? | **Reject `--`** | Enforce delimiter rule on all pools (existing and new) to avoid naming conflicts |
+| Drain monitoring interval | **15 min (configurable)** | Expose via workflow config field for tuning |
```
RyaliNvidia, line 446: Do you have a section explaining what this is?

projects/PROJ-220-dynamic-subpool.md
```diff
@@ -0,0 +1,944 @@
...
+| OSMO as single enforcement point? | **Yes** | KAI does not reserve idle children's quota; OSMO admission is the only defense |
+| Bypass protection (direct K8s)? | **Accepted risk** | Workloads bypassing OSMO may violate quotas; document as limitation |
+| Subpool API endpoints? | **Hybrid** | Writes use `/pool/{parent}/subpool` (enforce parent context, quota constraints, state semantics); reads use existing pool endpoint with canonical name (`/pool/{parent}--{subpool}`) since subpools are pool table entries |
```
RyaliNvidia, line 449: subpool are pool table entries, but they are separate columns for the parent and subpool, so the api would still need to split the string on the api side. So I think my previous comment might make more sense

projects/PROJ-220-dynamic-subpool.md
RyaliNvidia: Honestly `/` might be fine because you are separating pool and subpool into two different places in the api endpoint (If you change your GET api for subpool)

projects/PROJ-220-dynamic-subpool.md
```diff
@@ -0,0 +1,944 @@
...
+
+#### Display Changes (Visibility Only)
+
+- Parent "Available" may become negative during drain for pre-existing usage.
+- This is a visibility change (new column semantics) rather than a behavioral break.
+
+#### Migration Path
+
+No migration is required for existing pools:
+1. On deployment, OSMO will create `--_shared` queues for all existing pools
```
RyaliNvidia, line 503: Won't we need to trigger the creation of the shared pools for existing pools? How will this happen? Because by default, it doesn't try to recreate the queue if it is already using KAI cause we assume that the queue has already been created

projects/PROJ-220-dynamic-subpool.md
```diff
@@ -0,0 +1,944 @@
...
+### Operations
+
+- Requires a database migration to extend the `pools` table with subpool fields.
+- Requires operational visibility for subpool states (`ACTIVE/DELETING/ARCHIVED`) and drain progress.
+
+### Documentation
+
+- CLI docs for `osmo pool subpool ...`
+- Updated `osmo pool list` semantics (negative Available while draining)
+- Admin/operator runbook notes for delete/drain behavior
```
RyaliNvidia, line 521: We should probably have a runbook for expected behaviors and examples for how shared works and whatnot. Just for when we give the power to users to manage the subpools

projects/PROJ-220-dynamic-subpool.md
```diff
@@ -0,0 +1,944 @@
...
+7. **Tests & Docs**
+   - E2E tests for create/update/delete and admission behavior
+   - Verify acceptance criteria AC1-AC8
+   - Verify backwards compatibility for pools without subpools
+
+## Open Questions
+
+_None at this time._
+
+## Additional Considerations
```
RyaliNvidia, line 729: One thing I would like to see which isn't implemented yet is pool renaming. So subpool renaming is something I would like to also be implemented in the initial design as people constantly like renaming stuff

projects/PROJ-220-dynamic-subpool.md
RyaliNvidia: Seems like draining is when a pool is deleted. Do we need to drain the workflows? Can we leave the workflows to complete on their own, and if the users want them deleted, they can cancel them themselves?

projects/PROJ-220-dynamic-subpool.md
```diff
@@ -0,0 +1,944 @@
...
+osmo workflow submit --pool <parent--subpool> my-workflow.yaml
+```
+
+Example `osmo pool list` output (HIGH/NORMAL only; Available can be negative while draining):
+
+```
+Pool            Status         Subpool State  GPU Quota        Used  Available
+--------------  -------------  -------------  ---------------  ----  ---------
+team            ONLINE         -              10 (Total: 100)  50    -40
+├─ team--a      ONLINE         ACTIVE         30               5     25
```
RyaliNvidia, line 223: Do we need to show subpool state? Should we only show ACTIVE subpools here?

projects/PROJ-220-dynamic-subpool.md
RyaliNvidia: Actually, I didn't think about the workflow submit api which was the original reason we didn't want `/`

fernandol-nvidia: Yes, url safety is something we want, as well as k8s object names safety (https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#path-segment-names). We might need to stick to `--` for now.

projects/PROJ-220-dynamic-subpool.md
fernandol-nvidia: We will need to manage user expectation. I think subpools should inherit with very limited overrides (which we can add later). But being strict about this will make it easier to implement as well as operate in the wild.

projects/PROJ-220-dynamic-subpool.md
fernandol-nvidia: `{parent}--{subpool}` is the actual canonical name for the subpool in the pool table. So we can just look it up directly. But I can see the ergonomics behind `/{pool}/subpool/{subpool}`.

projects/PROJ-220-dynamic-subpool.md
fernandol-nvidia: Yup definitely, I will create one as we get closer to feature complete. Thanks!

projects/PROJ-220-dynamic-subpool.md
fernandol-nvidia: I agree that renaming is nice and should be something to support. I think the logic/mechanism would be similar for pool/subpool renaming (e.g. we need to update workflows/queues). And I think this would add quite a bit of scope to the project. If it makes sense, I would like to carve that out into a separate design and tackle renaming for both Pool + Subpool.

projects/PROJ-220-dynamic-subpool.md
fernandol-nvidia: In order to balance our GPU accounting, we still need to show DELETING, but we do not need to show ARCHIVED. Let me add a note.

projects/PROJ-220-dynamic-subpool.md
fernandol-nvidia: We will need to trigger the creation of the shared queue. But the existing queue will exist as well (and we cannot change its quota due to the nature of hierarchical queue). So existing workflows in the queue will still work. At OSMO, we need to prevent new jobs from going to the previous queue and strictly to shared or children queues.

projects/PROJ-220-dynamic-subpool.md
fernandol-nvidia: We will leave the workflows to complete on their own. The pool admin will need to decide if they want to force drain or natural drain. Added more details about the mechanism in [appendix](https://github.com/NVIDIA/OSMO/blob/2a77d824157333005f363c8547223b865c3ef4f1/projects/PROJ-220-dynamic-subpool.md).

projects/PROJ-220-dynamic-subpool.md
fernandol-nvidia: Yup, that makes sense. Updated

projects/PROJ-220-dynamic-subpool.md
fernandol-nvidia: Great call out, this is very tricky. Let me start a new dedicated PR to address this after I merge the initial design.

projects/PROJ-220-dynamic-subpool.md
```diff
@@ -0,0 +1,956 @@
...
+
+Solution: Create an explicit leaf queue ("shadow queue") for parent submissions
+
+Before (problematic):
+  osmo-default → osmo-pool-team (parent AND leaf, quota=100)
+                 └── Jobs submitted here can consume child quota
+
+After (shadow queue pattern):
+  osmo-default → osmo-pool-team (parent container, quota=100)
+                 ├── osmo-pool-team--_shared (leaf, quota=10)  ← parent submissions go here
```
ethany-nv, line 151: I wonder if we should have a default subpool, like how we have default pool. So that when people submit it just goes to a default subpool, and that 'shared' portion of the pool will just keep old workflows that will eventually be flushed out

RyaliNvidia: We can probably just support specifying subpool in the profile default pool section

RyaliNvidia (approved):

projects/PROJ-220-dynamic-subpool.md
fernandol-nvidia: +1, default subpool in profile sounds like a good addition. The follow up PR will be to discuss the guardrail re: the parent pool (IF we have a subpool). Some preliminary ideas are:

- parent pool becomes pre-emptible by sub-pools
- parent pool becomes blocked from submission and a new "flex" pool is created
- parent pool only accepts LOW

Issue #232, opened 2026-01-15
Construct base auth service which performs role check


RyaliNvidia:
The goal is to begin the auth rework by constructing the auth verification process using the API paths to mimic the current behavior.

The next step would be to convert the verification from evaluating API paths directly to the resource action model

PR #240, opened 2026-01-15
#232 - Add authz sidecar service with Go implementation

RyaliNvidia:
This commit adds the authorization sidecar service including:
- Go-based authz server implementing Envoy External Authorization
- PostgreSQL client for role/policy storage
- Role caching for performance optimization
- Action registry for path-to-action mapping
- Comprehensive test suite

Issue #232

## Checklist

### DIFF
--- a/src/service/authz_sidecar/main.go
+++ b/src/service/authz_sidecar/main.go
@@ -126,7 +126,12 @@ func main() {
 		slog.Int("max_size", *cacheMaxSize),
 	)
 
-	authzServer := server.NewAuthzServer(pgClient, roleCache, logger)
+	// Create role fetcher using the postgres client
+	roleFetcher := func(ctx context.Context, roleNames []string) ([]*postgres.Role, error) {
+		return postgres.GetRoles(ctx, pgClient, roleNames)
+	}
+
+	authzServer := server.NewAuthzServer(pgClient, roleFetcher, roleCache, logger)
 
 	logger.Info("authz server initialized")
 
--- a/src/utils/postgres/postgres_client.go
+++ b/src/utils/postgres/postgres_client.go
@@ -134,13 +134,6 @@ func NewPostgresClient(ctx context.Context, config PostgresConfig, logger *slog.
 	}, nil
 }
 
-// GetRoles retrieves roles by their names from the database.
-// This method wraps the standalone GetRoles function to satisfy the
-// PostgresClientInterface, which allows mock implementations in tests.
-func (c *PostgresClient) GetRoles(ctx context.Context, roleNames []string) ([]*Role, error) {
-	return GetRoles(ctx, c, roleNames)
-}
-
 // Close closes the database connection pool
 func (c *PostgresClient) Close() {
 	c.logger.Info("closing postgres client")
--- a/src/service/authz_sidecar/server/authz_server.go
+++ b/src/service/authz_sidecar/server/authz_server.go
@@ -45,25 +45,31 @@ const (
 
 // PostgresClientInterface defines the interface for PostgreSQL operations
 type PostgresClientInterface interface {
-	GetRoles(ctx context.Context, roleNames []string) ([]*postgres.Role, error)
 	Close()
 	Ping(ctx context.Context) error
 }
 
+// RoleFetcher is a function type for fetching roles from the database.
+// This allows the authz server to be decoupled from the postgres package,
+// enabling easier testing with mock implementations.
+type RoleFetcher func(ctx context.Context, roleNames []string) ([]*postgres.Role, error)
+
 // AuthzServer implements Envoy External Authorization service
 type AuthzServer struct {
 	envoy_service_auth_v3.UnimplementedAuthorizationServer
-	pgClient  PostgresClientInterface
-	roleCache *RoleCache
-	logger    *slog.Logger
+	pgClient    PostgresClientInterface
+	roleFetcher RoleFetcher
+	roleCache   *RoleCache
+	logger      *slog.Logger
 }
 
 // NewAuthzServer creates a new authorization server
-func NewAuthzServer(pgClient PostgresClientInterface, roleCache *RoleCache, logger *slog.Logger) *AuthzServer {
+func NewAuthzServer(pgClient PostgresClientInterface, roleFetcher RoleFetcher, roleCache *RoleCache, logger *slog.Logger) *AuthzServer {
 	return &AuthzServer{
-		pgClient:  pgClient,
-		roleCache: roleCache,
-		logger:    logger,
+		pgClient:    pgClient,
+		roleFetcher: roleFetcher,
+		roleCache:   roleCache,
+		logger:      logger,
 	}
 }
 
@@ -157,7 +163,7 @@ func (s *AuthzServer) checkAccess(ctx context.Context, path, method string, role
 	if !found {
 		// Query PostgreSQL
 		var err error
-		roles, err = s.pgClient.GetRoles(ctx, roleNames)
+		roles, err = s.roleFetcher(ctx, roleNames)
 		if err != nil {
 			return false, err
 		}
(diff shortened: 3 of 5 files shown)


src/service/authz_sidecar/integration_test.go
```diff
@@ -0,0 +1,224 @@
+/*
+SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
xutongNV: 2026

src/service/authz_sidecar/server/action_registry.go
```diff
@@ -0,0 +1,683 @@
...
+
+	// Extract resource type from action (e.g., "workflow:Create" -> "workflow")
+	actionParts := strings.Split(action, ":")
+	if len(actionParts) < 1 {
+		return "*"
+	}
+	resourceType := actionParts[0]
+
+	// Determine scope based on resource type (from Resource-Action Model)
+	switch resourceType {
```
fernandol-nvidia: Strongly type all resource types with enums please ([examples](https://gobyexample.com/enums))

src/service/authz_sidecar/server/role_cache.go
```diff
@@ -0,0 +1,236 @@
...
+	config  RoleCacheConfig
+	mu      sync.RWMutex
+	logger  *slog.Logger
+	evicted int64
+	hits    int64
+	misses  int64
+}
+
+// NewRoleCache creates a new role cache
+func NewRoleCache(config RoleCacheConfig, logger *slog.Logger) *RoleCache {
```
fernandol-nvidia, line 49: food for thought, there are existing lru cache for golang that's been hyper optimized and doesn't require us to maintain as much custom code (e.g. https://github.com/hashicorp/golang-lru, https://github.com/elastic/go-freelru)

src/utils/postgres/postgres_client.go
```diff
@@ -0,0 +1,226 @@
...
+}
+
+// PostgresClient handles PostgreSQL database operations
+type PostgresClient struct {
+	pool   *pgxpool.Pool
+	logger *slog.Logger
+}
+
+// RoleAction represents a single role action
+type RoleAction struct {
```
fernandol-nvidia: nit: perhaps coordinate with @xutongNV and see if we will have a centralized postgres utils for golang moving forward. 

I suspect we should keep the role/auth related stuff somewhere else because I think postgres will be one of the most highly used utils as we write more and more golang services.

src/service/authz_sidecar/BUILD
```diff
@@ -0,0 +1,228 @@
...
+        "//src/service/authz_sidecar/server:server",
+        "//src/utils/postgres:postgres",
+        "@org_golang_google_grpc//:go_default_library",
+        "@org_golang_google_grpc//health:go_default_library",
+        "@org_golang_google_grpc//health/grpc_health_v1:go_default_library",
+        "@org_golang_google_grpc//keepalive:go_default_library",
+    ],
+)
+
+go_binary(
```
xutongNV: you have authz_sidecar_bin_x86_64 and authz_sidecar_bin_arm below, is this one redundant?

src/service/authz_sidecar/main.go
```diff
@@ -0,0 +1,160 @@
...
+	cacheTTL     = flag.Duration("cache-ttl", defaultCacheTTL, "Cache TTL for roles")
+	cacheMaxSize = flag.Int("cache-max-size", defaultCacheSize, "Maximum cache size")
+)
+
+func main() {
+	flag.Parse()
+
+	// Setup structured logging
+	logger := slog.New(slog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{
+		Level: slog.LevelInfo,
```
xutongNV: can this be configured?

src/service/authz_sidecar/main.go
```diff
@@ -0,0 +1,160 @@
...
+	opts := []grpc.ServerOption{
+		grpc.KeepaliveParams(keepalive.ServerParameters{
+			Time:    60 * time.Second,
+			Timeout: 20 * time.Second,
+		}),
+		grpc.KeepaliveEnforcementPolicy(keepalive.EnforcementPolicy{
+			MinTime:             30 * time.Second,
+			PermitWithoutStream: true,
+		}),
+		grpc.MaxRecvMsgSize(4 * 1024 * 1024), // 4MB
```
xutongNV: can you move this to the top of the file as constants

src/utils/postgres/postgres_client.go
xutongNV: Yeah maybe we could have a centralize postgres connector. 
https://github.com/NVIDIA/OSMO/blob/feature/PROJ-147-operator-redesign/src/service/operator/main.go#L81

src/utils/postgres/postgres_client.go
```diff
@@ -0,0 +1,226 @@
...
+	Base   string `json:"base"`
+	Path   string `json:"path"`
+	Method string `json:"method"`
+}
+
+// RolePolicy represents a role policy with multiple actions
+type RolePolicy struct {
```
xutongNV: In python we have everything that uses postgres in the postgres file and redis likewise. But this makes it difficult to find the code that you are looking for. 
Since we are staring fresh in Go, I would suggest only keeping connection related stuff in post_gres.go and having role related codes into a role.go or role/

src/utils/postgres/postgres_client.go
```diff
@@ -0,0 +1,226 @@
...
+	logger.Info("postgres client connected successfully")
+
+	return &PostgresClient{
+		pool:   pool,
+		logger: logger,
+	}, nil
+}
+
+// GetRoles retrieves roles by their names from the database
+func (c *PostgresClient) GetRoles(ctx context.Context, roleNames []string) ([]*Role, error) {
```
xutongNV: this should be a method of Role, not PostgresClient

src/service/authz_sidecar/server/action_registry.go
RyaliNvidia: Sounds good. Done

src/service/authz_sidecar/BUILD
RyaliNvidia: Yeah. This was used for testing purposes, but the x86 one can be used instead

src/utils/postgres/postgres_client.go
RyaliNvidia: Yeah makes sense. I'll move it to a separate file for now

src/utils/postgres/postgres_client.go
RyaliNvidia: I'll move the logic into a separate file and have this function call that one, but this function is used for the mock testing for the postgres client.

src/utils/postgres/postgres_client.go
RyaliNvidia: Yeah sounds good. Consolidated some of the postgres stuff and removed the Role stuff

src/service/authz_sidecar/server/role_cache.go
RyaliNvidia: Yeah. This is a better idea

src/service/authz_sidecar/BUILD
xutongNV: nit: can we remove this?

src/utils/postgres/postgres_client.go
elookpotts-nvidia: Agreed that this should be a method of Role, not PostgresClient. We shouldn't maintain implementation just for testing - that's indicative of testing the PostgresClient improperly.

The PostgresClient will balloon with business logic quickly, just like our python PostgresClient. Instead, we can have the Role struct accept a PostgresClient. This would be more functional like Go, as opposed to object-oriented like Python.

src/service/authz_sidecar/main.go
elookpotts-nvidia, line 1: Should the performance comparison test be checked in?

RyaliNvidia: I think the test is still worth keeping in so we can keep evaluating it as we go since more changes will come to the logic

src/utils/postgres/postgres_client.go
RyaliNvidia: It's not for the PostgresClient testing but the full auth service testing. I'll remove it out of the postgres client anyways

Issue #244, opened 2026-01-16
Relogin should remember the last used login url


RyaliNvidia:
### What new feature or improvement would you like to see?

Rather than typing in the same login url every time the user runs `osmo login`, since the last url is saved, it should fetch it by default.

### Why is this needed?

Nice quality of life improvement

### Code of Conduct

- [x] I agree to follow NVIDIA OSMO's Code of Conduct
- [x] I have searched the [open feature requests](https://github.com/NVIDIA/OSMO/issues?q=is%3Aopen+is%3Aissue+label%3Afeature) and have found no duplicates for this feature request

PR #245, opened 2026-01-16
#244 - Use last login url if url is not specified

RyaliNvidia:
## Description
If the user enters just `osmo login`, it should use the last url that was used for logging in

Also added a message about what url is being used:
```
$ osmo login
Logging in to url: <url>
...
```

Issue #244

## Checklist

### DIFF
--- a/src/cli/login.py
+++ b/src/cli/login.py
@@ -1,5 +1,6 @@
 """
-SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES.
+All rights reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -76,10 +77,10 @@ def _login(service_client: client.ServiceClient, args: argparse.Namespace):
     # Get the url from args or fall back to last used url
     try:
         url = args.url or service_client.login_manager.login_storage.url
-    except osmo_errors.OSMOUserError:
+    except osmo_errors.OSMOUserError as error:
         raise osmo_errors.OSMOUserError(
             'No url provided and no previous login found. '
-            'Please provide a url: osmo login <url>')
+            'Please provide a url: osmo login <url>') from error
 
     # Validate the url
     class UrlValidator(pydantic.BaseModel):


src/cli/login.py
```diff
@@ -71,13 +73,13 @@ def setup_parser(parser: argparse._SubParsersAction):
...
 def _login(service_client: client.ServiceClient, args: argparse.Namespace):
-    # Validate the url
-    class UrlValidator(pydantic.BaseModel):
-        url: pydantic.AnyHttpUrl
+    # Get the url from args or fall back to last used url
     try:
-        _ = UrlValidator(url=args.url)
-    except pydantic.error_wrappers.ValidationError as error:
-        raise osmo_errors.OSMOUserError(f'Bad url {args.url}: {error}')
+        url = args.url or service_client.login_manager.login_storage.url
```
xutongNV, line 79: maybe it is good to have a log message here to show the url that will be used or even let user confirm before call login function?

RyaliNvidia: Printing out a comment of the url is a good idea

I would say 99% of the time, the user will have to confirm anyway since they will need to authenticate against their auth provider so I feel like a confirmation might add an additional unneeded level of friction

xutongNV (approved):

src/cli/login.py
```diff
@@ -71,13 +73,15 @@ def setup_parser(parser: argparse._SubParsersAction):
...
+    except osmo_errors.OSMOUserError:
+        raise osmo_errors.OSMOUserError(
+            'No url provided and no previous login found. '
+            'Please provide a url: osmo login <url>')
+
+    print(f'Logging in to url: {url}')
```
ethany-nv:
```suggestion
    print(f'Logging in to {url}')
```

src/cli/login.py
```diff
@@ -71,13 +73,15 @@ def setup_parser(parser: argparse._SubParsersAction):
...
 def _login(service_client: client.ServiceClient, args: argparse.Namespace):
-    # Validate the url
-    class UrlValidator(pydantic.BaseModel):
-        url: pydantic.AnyHttpUrl
+    # Get the url from args or fall back to last used url
     try:
-        _ = UrlValidator(url=args.url)
-    except pydantic.error_wrappers.ValidationError as error:
-        raise osmo_errors.OSMOUserError(f'Bad url {args.url}: {error}')
+        url = args.url or service_client.login_manager.login_storage.url
```
ethany-nv, line 79: Should we feed this url variable to the UrlValidator?

RyaliNvidia: Yeah. I'll add it back

xutongNV (approved):

PR #251, opened 2026-01-17
Database Pooling in Postgres Singleton Across Services

ethany-nv:
## Description
This PR introduces database pooling, which allows the PostgresConnector singleton to instantiate multiple database connections at once. This way, each service instance can have multiple connections available for the threads/coroutines that are created when users make API requests, instead of coroutines having to wait on just one connection before their database queries go through.

For example, when there is a surge of users submitting workflows, or when the service is redeployed and all the workflows are making auth requests, the service will be able to handle the requests more efficiently.

Two new parameters are introduced:
```
postgres_pool_minconn: Minimum number of connections to keep in the connection pool
postgres_pool_maxconn: Maximum number of connections to keep in the connection pool
```
Issue #262 

## Checklist

### DIFF
--- a/src/utils/connectors/postgres.py
+++ b/src/utils/connectors/postgres.py
@@ -47,7 +47,7 @@ from starlette.types import ASGIApp, Receive, Scope, Send
 from src.lib.data import storage
 from src.lib.data.storage import constants
 from src.lib.utils import (common, credentials, jinja_sandbox, login,
-                           osmo_errors, role)
+                           osmo_errors, role, validation)
 from src.utils import auth, notify
 from src.utils.secret_manager import Encrypted, SecretManager
 
@@ -155,6 +155,7 @@ class PostgresConfig(pydantic.BaseModel):
     postgres_reconnect_retry: int = pydantic.Field(
         command_line='postgres_reconnect_retry',
         env='OSMO_POSTGRES_RECONNECT_RETRY',
+        type=validation.positive_integer,
         default=5,
         description='Reconnect try count after connection error')
     mek_file: str = pydantic.Field(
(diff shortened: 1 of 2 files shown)


src/utils/connectors/postgres.py
```diff
@@ -268,9 +386,23 @@ def __init__(self, config: PostgresConfig):
         self._init_configs()
         logging.debug('Configs initialized')
 
+        # Register cleanup on exit
+        atexit.register(self.close)
```
RyaliNvidia, line 391: Is this needed? I'm not sure what this does

ethany-nv: This is the standard for handling connection cleanup when the service exits. This has better guarantees for cleaning up the connection when something goes wrong - python will call self.close at the very end no matter what

src/utils/connectors/postgres.py
```diff
@@ -296,38 +428,39 @@ def execute_fetch_command(self, command: str,
...
+                    rows = [
+                        pydantic.create_model(
+                            'DynamicModel', **{k: common.handle_memoryview(v) or \
+                                               (Any, common.handle_memoryview(v))
+                                               for k, v in row.items()})()  # type: ignore
+                        for row in rows]
+                cur.close()
+                conn.commit()
+                return rows
+            except (psycopg2.DatabaseError, psycopg2.InterfaceError) as error:
```
RyaliNvidia, line 451: Can all these get pool errors as well?

src/utils/connectors/postgres.py
RyaliNvidia: Got it

xutongNV (dismissed):

src/utils/connectors/postgres.py
ethany-nv: If there is a pool error, this will be caught in the retry decorator, and the decorator will handle it.

RyaliNvidia (approved):

PR #261, opened 2026-01-20
Add task status to workflow listener

xutongNV:
## Description

Compute task status from pod

Issue #147

## Checklist

### DIFF
--- a/src/operator/utils/BUILD
+++ b/src/operator/utils/BUILD
@@ -63,16 +63,17 @@ go_library(
 )
 
 # Run this  manually
-# go_test(
-#     name = "calculate_status_test",
-#     srcs = ["calculate_status_test.go"],
-#     data = ["//src/operator/tests:test_calculate_pod_status_cases.json"],
-#     embed = [":utils"],
-#     deps = [
-#         "@io_k8s_api//core/v1:core",
-#         "@io_k8s_apimachinery//pkg/apis/meta/v1:meta",
-#     ],
-# )
+go_test(
+    name = "calculate_status_test",
+    srcs = ["calculate_status_test.go"],
+    data = ["//src/operator/tests:test_calculate_pod_status_cases.json"],
+    embed = [":utils"],
+    tags = ["manual"],
+    deps = [
+        "@io_k8s_api//core/v1:core",
+        "@io_k8s_apimachinery//pkg/apis/meta/v1:meta",
+    ],
+)
 
 go_test(
     name = "unack_messages_test",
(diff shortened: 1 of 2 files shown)


src/operator/utils/container_status.go
```diff
@@ -0,0 +1,245 @@
...
+		"ErrImagePull":               302,
+		"CreateContainerConfigError": 303,
+		"CrashLoopBackOff":           304,
+		"ContainerStatusUnknown":     305,
+	}
+)
+
+// ContainerStatus holds container reasons and exit codes
+type ContainerStatus struct {
+	Reasons   map[string]string // container name -> reason (waiting or terminated)
```
fernandol-nvidia, line 62: Is it possible for reasons to be an enum? ([example](https://gobyexample.com/enums))

fernandol-nvidia: Oh i just realized we are grabbing this directly from k8s's `*corev1.Pod`... I guess they are returning raw strings, so prob not worth creating wrappers around them.

src/operator/utils/BUILD
```diff
@@ -59,3 +61,26 @@ go_library(
         "@io_k8s_client_go//tools/clientcmd",
     ],
 )
+
+# Run this  manually
+# go_test(
```
fernandol-nvidia: nit: we can uncomment and add tags = [ "manual" ]

src/operator/utils/container_status.go
```diff
@@ -0,0 +1,245 @@
...
+	"slices"
+	"strings"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+// Container exit code offsets
+var containerExitCodeOffsets = map[string]int32{
+	"osmo-init":      255,  // ExitCodeOffsetInit
```
fernandol-nvidia: This can be [strongly typed](https://gobyexample.com/enums) since we control this.

src/operator/utils/task_status.go
```diff
@@ -0,0 +1,201 @@
...
+	Status   string
+	Message  string
+	ExitCode int32
+}
+
+// getPodConditionFailure checks pod conditions for failures (preemption, disruption, etc.)
+// Returns the TaskStatusResult if a failure condition is found, nil otherwise.
+func getPodConditionFailure(pod *corev1.Pod) *TaskStatusResult {
+	for _, cond := range pod.Status.Conditions {
+		if cond.Status == corev1.ConditionTrue {
```
fernandol-nvidia, line 63: what happens if neither of the below conditions match? We skip over this condition?

xutongNV: Yes. If no special conditions, we check container status and pod phase

PR #274, opened 2026-01-21
Add message worker to chart

xutongNV:
## Description

rename agent worker to message worker
add message worker to all scripts of launching osmo service

Issue #147 

## Checklist

### DIFF
--- a/run/start_service_bazel.py
+++ b/run/start_service_bazel.py
@@ -457,8 +457,8 @@ def start_service_bazel():
     check_required_tools(['bazel', 'docker', 'npm', 'aws'])
 
     try:
-        # _start_redis()
-        # _start_postgres()
+        _start_redis()
+        _start_postgres()
         _start_localstack_s3()
         _create_localstack_buckets()
 
--- a/src/service/agent/message_worker.py
+++ b/src/service/agent/message_worker.py
@@ -44,8 +44,8 @@ MESSAGE_CLAIM_IDLE_TIME_MS = 300000
 
 
 class MessageWorkerConfig(static_config.StaticConfig, connectors.RedisConfig,
-                         connectors.PostgresConfig, src.lib.utils.logging.LoggingConfig,
-                         metrics.MetricsCreatorConfig):
+                          connectors.PostgresConfig, src.lib.utils.logging.LoggingConfig,
+                          metrics.MetricsCreatorConfig):
     """Configuration for the message worker."""
     progress_file: str = pydantic.Field(
         command_line='progress_file',


deployments/charts/quick-start/values.yaml
```diff
@@ -466,6 +466,49 @@ service:
       - name: AWS_DEFAULT_REGION
         value: us-east-1
 
+    ## Message worker service configuration for processing backend messages
```
xutongNV, line 469: update all readmes

deployments/charts/service/values.yaml
```diff
@@ -916,6 +916,119 @@ services:
...
+    topologySpreadConstraints:
+    ## Spread message worker pods across different availability zones
+    ##
+    - topologyKey: topology.kubernetes.io/zone
+      maxSkew: 1
+      whenUnsatisfiable: ScheduleAnyway
+
+    ## Progress file path for liveness/startup probes
+    ##
+    progressFile: /tmp/osmo/service/last_progress_message_worker
```
vvnpn-nv, line 980: will this path ever change? it will stay the same unless user build images a different way right?

vvnpn-nv: if it's always static maybe we keep it static in the chart as well, instead of making it configurable

ethany-nv: I think this is a good point. There is never a reason to change this

xutongNV: No. I dont expect it to be change right now. However it did change in the past and because we dont have this configurable I had to workaround in the codes to change to from /var/run/osmo to /tmp/osmo.
So I fugured we should make this flexible just in case

run/start_service_bazel.py
```diff
@@ -429,18 +429,42 @@ def _start_router_service():
...
+
+
 def start_service_bazel():
     """Start the OSMO service using bazel."""
     check_required_tools(['bazel', 'docker', 'npm', 'aws'])
 
     try:
-        _start_redis()
-        _start_postgres()
+        # _start_redis()
```
ethany-nv: Remove comment?

src/service/agent/message_worker.py
```diff
@@ -37,20 +37,20 @@
...
 # Time in milliseconds before a pending message is considered abandoned (5 minutes)
 MESSAGE_CLAIM_IDLE_TIME_MS = 300000
 
 
-class AgentWorkerConfig(static_config.StaticConfig, connectors.RedisConfig,
-                        connectors.PostgresConfig, src.lib.utils.logging.LoggingConfig,
-                        metrics.MetricsCreatorConfig):
-    """Configuration for the agent worker."""
+class MessageWorkerConfig(static_config.StaticConfig, connectors.RedisConfig,
+                         connectors.PostgresConfig, src.lib.utils.logging.LoggingConfig,
```
ethany-nv: nit: line up the parameters with previous line

Issue #237, opened 2026-01-15
Config Edit confirmation UI


patclarknvidia:
Piggy-back on "Highlight diff between 2 service configs" to show the changes that are made in the config editor and allow confirmation before save.

Issue #236, opened 2026-01-15
Create generic MultiValEdit control


patclarknvidia:
For fields like user roles and Control roles we need a control that allows showing the current tags and adding/deleting.

Issue #234, opened 2026-01-15
Edit Service Config UI page


patclarknvidia:
UI should be similar to the service config details modal which flattens the JSON that is divided into unnecessary sections like cli_config and login_info.

PR #277, opened 2026-01-22
Service Config History and Edit

patclarknvidia:
See /admin/service and /admin/service/history

- Display history of service config changes table
- Display service config details modal
- Highlight diff between 2 service configs
- Edit Service Config UI page
- Create generic MultiValEdit control
- Config Edit confirmation UI

Issue #226 
Issue #227
Issue #228
Issue #234 
Issue #236
Issue #237

## Checklist

image
image
image
image

### DIFF
--- a/ui/src/app/admin/service/history/page.tsx
+++ b/ui/src/app/admin/service/history/page.tsx
@@ -64,7 +64,7 @@ export default function AdminPage() {
   return (
     <HistoryDetailsProvider configs={configHistory.data?.configs}>
       <PageHeader title="Service Config History">
-        {selectedRevisions.length > 1 && (
+        {selectedRevisions.length === 2 && (
           <IconButton
             className="btn btn-primary"
             icon="compare"


ui/src/app/admin/service/components/ServiceConfigEditor.tsx
```diff
@@ -0,0 +1,550 @@
...
+
+interface ServiceConfigEditorProps {
+  serviceConfig: ServiceConfig;
+  onSave: (description: string, tags: string[], config: ServiceConfig) => void;
+  error?: string;
+}
+
+type ServiceConfigFormValues = yup.InferType<typeof serviceConfigSchema>;
+
+export const ServiceConfigEditor = ({ serviceConfig, onSave, error }: ServiceConfigEditorProps) => {
```
elookpotts-nvidia, line 41: Hey, this is looking pretty great. Nice stuff.

One concern: the `ServiceConfigEditor` component and the `/admin/service/` route are pretty tightly coupled to the specific shape of `ServiceConfig`. Looking at `ServiceConfigEditor.tsx`, every field is manually mapped in `defaultValues`, laid out individually in the TSX, and then manually reconstructed in `onSubmit`. Same story with `ServiceConfigSchema.tsx` -- there's a yup schema that mirrors the zod schema in `models/config/service-config.ts`, but with hand-written validation for each field. IMO this will become a maintenance headache when we add more config types. Ideally:

1. The route can be dynamic: `/admin/config/[configType]/` as a slug route rather than `/admin/service/`. Pass the config type as a param.

2. The UI can be derived from the schema: The zod schemas in `models/config/service-config.ts` already define the structure. You could use those definitions (or extend them with additional metadata like labels, input types, validation hints) to dynamically render form fields. There are libraries for this, or you could build a lightweight renderer that walks the schema.

3. Only use zod for validation: Instead of maintaining a separate yup schema that duplicates the zod definitions, consider using `zod` for form validation directly (e.g., `@hookform/resolvers/zod`) or generating one from the other.

The goal is that when someone adds a new config type like backend or pod template configs they define it once in `models/config/`, and the UI just works without needing a new page and another massive editor component.

Totally get that a lot of work went into this, and I'll approve so you can get it merged. These improvements can definitely be done in a follow-up PR (yay feature branches for in-progress projects!)

patclarknvidia: Agree - it's easier to come up with a framework once I'm on the second config and I will look to make it more generic then. Hard to make it generic when you're just doing the first one.

PR #281, opened 2026-01-23
Resource Listener

ethany-nv:
## Description

This re-implementation of resource listening uses a more efficient way to aggregate resource usage:

Pod resource usage is computed incrementally from informer deltas: when a pod first transitions into Running (with a node assignment), its requested resources are added to per‑node totals; when it transitions out to a terminal phase or is deleted, its contribution is subtracted. Because pod requests and node placement are immutable, repeated updates for the same UID are ignored. Contributions are derived from container requests (CPU millicores, memory/ephemeral storage Ki, GPUs) and tracked both as overall usage and non‑workflow usage; dirty nodes are flushed on a debounce interval, and any watch gaps trigger a full rebuild from the informer cache.

Issue #329 

## Checklist

### DIFF
--- a/src/operator/backend_listener.py
+++ b/src/operator/backend_listener.py
@@ -627,7 +627,7 @@ def update_resource_usage(node_send_queue: helpers.EnqueueCallback,
     non_wf_resource_usage = format_resource_usage(non_wf_requests)
 
     resource_message = backend_messages.MessageBody(
-        type=backend_messages.MessageType.RESOURCE_USAGE,
+        type=backend_messages.MessageType.UPDATE_NODE_USAGE,
         body=backend_messages.UpdateNodeUsageBody(
             hostname=node_name,
             usage_fields=resource_usage,


src/operator/tests/node_helpers_test.go
```diff
@@ -0,0 +1,644 @@
...
+func TestBuildResourceBody(t *testing.T) {
+	timeNow := metav1.NewTime(time.Now())
+
+	tests := []struct {
+		name     string
+		node     *corev1.Node
+		isDelete bool
+		validate func(*testing.T, interface{})
+	}{
+		{
```
xutongNV: should load test cases from the json

src/operator/resource_listener.go
```diff
@@ -0,0 +1,633 @@
...
+	)
+}
+
+// receiveMessages handles receiving ACK messages from the server
+// sendMessages starts informers and processes resource events
+func (rl *ResourceListener) sendMessages() {
+	// Create channels for different message types
+	// resourceChan: node resource messages (from watchResources)
+	// usageChan: pod resource usage messages (from watchPods)
+	resourceChan := make(chan *pb.ListenerMessage, rl.args.PodUpdateChanSize)
```
xutongNV: rename resource to node

src/operator/utils/listener_args.go
```diff
@@ -52,6 +56,12 @@ func ListenerParse() ListenerArgs {
 	podUpdateChanSize := flag.Int("podUpdateChanSize",
 		getEnvInt("POD_UPDATE_CHAN_SIZE", 500),
 		"Buffer size for pod update channel")
```
xutongNV: can you add (WokflowListener) to the message

src/operator/BUILD
xutongNV: Go Listener

src/operator/BUILD
```diff
@@ -303,6 +303,7 @@ go_library(
     name = "workflow_listener_lib",
```
xutongNV: can you rename the target to listener_lib?

src/operator/BUILD
xutongNV: change target name, below as well

src/operator/resource_listener.go
```diff
@@ -0,0 +1,614 @@
...
+
+	// Start usage message sender goroutine (dedicated to usageChan)
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		rl.sendFromChannel("usage", usageChan, podWatcherDone, streamCtx, streamCancel)
+	}()
+
+	// Start progress reporter goroutine
+	wg.Add(1)
```
xutongNV, line 104: we should avoid have a coroutine that only does report progress

src/operator/resource_listener.go
```diff
@@ -0,0 +1,614 @@
...
+		defer wg.Done()
+		defer close(podWatcherDone)
+		rl.watchPods(usageChan)
+	}()
+
+	// Start resource message sender goroutine (dedicated to nodeChan)
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		rl.sendFromChannel("resource", nodeChan, nodeWatcherDone, streamCtx, streamCancel)
```
xutongNV: I wonder if watchPods and watchNodes can write to the same channel? Then we only need one send corountine

src/operator/resource_listener.go
```diff
@@ -0,0 +1,614 @@
...
+// sendFromChannel sends messages from a channel to the server
+// Each channel has its own dedicated sender to avoid starvation
+func (rl *ResourceListener) sendFromChannel(name string, msgChan <-chan *pb.ListenerMessage, watcherDone <-chan struct{}, streamCtx context.Context, streamCancel context.CancelCauseFunc) {
+	log.Printf("Starting %s message sender", name)
+	defer log.Printf("Stopping %s message sender", name)
+
+	for {
+		select {
+		case <-streamCtx.Done():
+			log.Printf("Context done, draining %s channel...", name)
```
xutongNV: draining can be removed from log

src/operator/resource_listener.go
```diff
@@ -0,0 +1,614 @@
...
+			return
+		case <-watcherDone:
+			log.Printf("%s watcher stopped unexpectedly, draining channel...", name)
```
xutongNV: same here

src/operator/resource_listener.go
```diff
@@ -0,0 +1,614 @@
...
+
+	if err := rl.GetStream().Send(msg); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// closeStream ensures stream is closed only once
+func (rl *ResourceListener) closeStream() {
```
xutongNV: this can be in base_listener

src/operator/resource_listener.go
```diff
@@ -0,0 +1,614 @@
...
+		if stream != nil {
+			if err := stream.CloseSend(); err != nil {
+				log.Printf("Error closing resource stream: %v", err)
+			}
+		}
+	})
+}
+
+// Close cleans up resources
+func (rl *ResourceListener) Close() {
```
xutongNV: this can be moved to base listener

src/operator/resource_listener.go
```diff
@@ -0,0 +1,582 @@
...
+			// Report progress periodically even when idle
+			progressWriter := rl.GetProgressWriter()
+			if progressWriter != nil {
+				if err := progressWriter.ReportProgress(); err != nil {
+					log.Printf("Warning: failed to report progress: %v", err)
+				}
+			}
+		case msg := <-nodeChan:
+			if err := rl.sendResourceMessage(msg); err != nil {
+				streamCancel(fmt.Errorf("failed to send resource message: %w", err))
```
xutongNV: node message

src/operator/resource_listener.go
```diff
@@ -0,0 +1,582 @@
...
+		defer wg.Done()
+		rl.sendFromChannels(nodeChan, usageChan, nodeWatcherDone, podWatcherDone, streamCtx, streamCancel)
+	}()
+
+	// Wait for all goroutines to complete
+	wg.Wait()
+	log.Println("All message sender goroutines stopped")
+}
+
+// sendFromChannels sends messages from both resource and usage channels to the server
```
xutongNV: node and usage channels

src/operator/utils/node_helpers.go
```diff
@@ -0,0 +1,115 @@
...
+// GetNodeHostname extracts the hostname from a node
+func GetNodeHostname(node *corev1.Node) string {
+	if hostname, ok := node.Labels["kubernetes.io/hostname"]; ok {
+		return hostname
+	}
+	return "-"
+}
+
+// BuildResourceBody creates an UpdateNodeBody from a node
+func BuildResourceBody(node *corev1.Node, isDelete bool) *pb.UpdateNodeBody {
```
xutongNV: BuildUpdateBody
Can you go through the changes and see if there are any places not renamed from resource to node?

src/operator/resource_listener.go
```diff
@@ -0,0 +1,582 @@
...
+				log.Printf("Sent resource_usage for node: %s", hostname)
+			case <-rl.GetStreamContext().Done():
+				return
+			}
+		}
+	}
+}
+
+// buildResourceMessage creates a ResourceBody message from a node
+func (rl *ResourceListener) buildResourceMessage(
```
xutongNV, line 496: can you call thie node message? It is a little confusing.
Also
```
  oneof body {
    UpdatePodBody update_pod = 3;
    LoggingBody logging = 4;
    UpdateNodeBody resource = 5;
    UpdateNodeUsageBody resource_usage = 6;
  }
```
can you align the name of this
```
    UpdateNodeBody update_node = 5;
    UpdateResourceUsageBody resource_usage = 6;
```

src/operator/resource_listener.go
```diff
@@ -0,0 +1,582 @@
...
+
+	// Process all nodes and send resource messages
+	nodes := nodeInformer.GetStore().List()
+	for _, obj := range nodes {
+		node, ok := obj.(*corev1.Node)
+		if !ok {
+			continue
+		}
+		// Force send for all nodes after rebuild
+		msg := rl.buildResourceMessageForced(node, nodeStateTracker)
```
xutongNV: nit: instead of have a forced func, you can do 
nodeStateTracker.empty()
buildResourceMessage()

src/operator/resource_listener.go
```diff
@@ -0,0 +1,582 @@
...
+	// We only care about:
+	// 1. Pods transitioning INTO Running state (to add resources)
+	// 2. Pods transitioning FROM Running state to terminal state (to subtract resources)
+	// All other updates (labels, annotations, status changes) are ignored since
+	// pod resources and node assignment are immutable after creation.
+	_, err = podInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			pod := obj.(*corev1.Pod)
+			// Only track if pod is already Running and has a node assignment
+			if pod.Spec.NodeName == "" || pod.Status.Phase != corev1.PodRunning {
```
xutongNV: In line 307 if you filtered pods with running phase, do you still need this check?
also if pod is running, can the node name be ""?

src/operator/resource_listener.go
```diff
@@ -0,0 +1,582 @@
...
+
+			// Handle two cases:
+			// Case 1: Pod transitioned TO Running state (Pending/Unknown → Running)
+			// Case 2: Pod transitioned FROM Running to terminal state (Running → Succeeded/Failed)
+			wasRunning := oldPod.Status.Phase == corev1.PodRunning
+			isRunning := newPod.Status.Phase == corev1.PodRunning
+
+			// Case 1: Pod just transitioned TO Running - track its resources
+			if !wasRunning && isRunning {
+				rl.aggregator.UpdatePod(newPod)
```
xutongNV: nit: aggregator.UpdatePod -> aggregator.AddPod to be consistent with "DeletePod"

src/operator/utils/node_usage_aggregator.go
```diff
@@ -0,0 +1,361 @@
...
+	rua.mu.Lock()
+	defer rua.mu.Unlock()
+
+	uid := pod.UID
+	oldContrib := rua.podContributions[uid]
+	if oldContrib == nil {
+		return
+	}
+
+	// Subtract contribution
```
xutongNV, line 243: should we check if the uid does exist, then subtract and remove?

src/operator/resource_listener.go
```diff
@@ -0,0 +1,582 @@
...
+			return
+		case <-func() <-chan time.Time {
+			if reconcileTicker != nil {
+				return reconcileTicker.C
+			}
+			return nil
+		}():
+			// Periodic full reconcile as safety net (only if enabled)
+			log.Println("Performing periodic node reconciliation...")
+			rl.rebuildNodesFromStore(nodeInformer, nodeStateTracker, nodeChan)
```
xutongNV: why do you need to resync nodes? I think informer already do that if you set
```
informerFactory := informers.NewSharedInformerFactoryWithOptions(
		clientset,
		time.Duration(args.ResyncPeriodSec)*time.Second,
         ...
	)
```

src/operator/utils/node_usage_aggregator.go
ethany-nv: In line 220, there is a check to see if the uid exists in the aggregator

src/operator/resource_listener.go
ethany-nv: Ok getting rid of reconciliation logic

src/operator/tests/BUILD
```diff
@@ -47,3 +52,16 @@ py_test(
         "//src/utils/job",
     ],
 )
+
+go_test(
+    name = "node_helpers_test",
+    srcs = ["node_helpers_test.go"],
+    data = ["test_node_helpers_cases.json"],
```
xutongNV, line 59: "test_node_usage_cases.json" is this used any where

src/operator/resource_listener.go
```diff
@@ -0,0 +1,526 @@
...
+	log.Println("Starting node resource watcher")
+
+	// State tracker for node deduplication
+	nodeStateTracker := utils.NewNodeStateTracker(time.Duration(rl.args.StateCacheTTLMin) * time.Minute)
+
+	// Create informer factory for nodes (cluster-scoped)
+	// Disable informer resync - rely on watch + error handlers
+	nodeInformerFactory := informers.NewSharedInformerFactory(
+		clientset,
+		0, // No automatic resync
```
xutongNV, line 217: should we add resync?

src/operator/resource_listener.go
```diff
@@ -0,0 +1,526 @@
...
+	}
+
+	log.Printf("Starting pod watcher for namespace: %s", rl.args.Namespace)
+
+	// Create informer factory for pods (all namespaces)
+	// Disable informer resync - rely on watch + error handlers
+	// Field selector for Running pods only to reduce memory footprint
+	podInformerFactory := informers.NewSharedInformerFactoryWithOptions(
+		clientset,
+		0, // No automatic resync
```
xutongNV, line 316: same here

src/operator/resource_listener.go
ethany-nv: I feel like there is no strong reason for adding resync, because if the watch connection doesn't error out, then our state in-memory should be up to date. If the connection breaks, we rebuild the state.

src/operator/utils/node_usage_aggregator.go
```diff
@@ -0,0 +1,383 @@
...
+
+// NodeUsageAggregator tracks per-node usage from pods
+type NodeUsageAggregator struct {
+	mu sync.RWMutex
+
+	// Workflow namespace for distinguishing workflow vs non-workflow pods
+	workflowNamespace string
+
+	// Pod contributions by UID
+	podContributions map[types.UID]*PodContribution
```
fernandol-nvidia, line 144: Check out the sync.Map package which has built in sync for things like (setDefault, LoadOrStore, etc). Also allows us to do granular locking per map.

src/operator/utils/node_usage_aggregator.go
```diff
@@ -0,0 +1,383 @@
...
+		if b[k] != v {
+			return false
+		}
+	}
+	return true
+}
+
+// NodeUsageAggregator tracks per-node usage from pods
+type NodeUsageAggregator struct {
+	mu sync.RWMutex
```
fernandol-nvidia, line 138: We have a coarse lock for everything... Should we? Different events might want to write to different maps in the aggregator that don't map to the same key. But coarse lock will cause everything to be written sequentially.

src/operator/utils/base_listener.go
```diff
@@ -219,3 +241,56 @@ func (bl *BaseListener) AddToWaitGroup(delta int) {
...
+	streamName string,
+) error {
+	// Initialize the base connection
+	if err := bl.InitConnection(ctx, bl.args.ServiceURL); err != nil {
+		return err
+	}
+
+	// Establish the bidirectional stream
+	var err error
+	bl.stream, err = bl.client.ListenerStream(ctx)
```
fernandol-nvidia: Could run be called multiple times? Do we need to lock or add once to the initialization?

src/operator/main.go
```diff
@@ -51,32 +60,71 @@ func main() {
 		log.Fatalf("Failed to initialize backend: %v", err)
 	}
 
+	// Create both listeners
 	workflowListener := NewWorkflowListener(cmdArgs)
+	resourceListener := NewResourceListener(cmdArgs)
+
+	// Set up signal handling for graceful shutdown
+	sigChan := make(chan os.Signal, 1)
+	signal.Notify(sigChan, os.Interrupt, syscall.SIGTERM)
```
fernandol-nvidia: Do we need this? doesn't context.Background() take care of this for us?

src/operator/resource_listener.go
```diff
@@ -0,0 +1,526 @@
...
+	// nodeChan: node resource messages (from watchNodes)
+	// usageChan: pod resource usage messages (from watchPods)
+	nodeChan := make(chan *pb.ListenerMessage, rl.args.NodeUpdateChanSize)
+	usageChan := make(chan *pb.ListenerMessage, rl.args.UsageChanSize)
+
+	// WaitGroup to track all goroutines
+	var wg sync.WaitGroup
+
+	// Channels to signal if watchers exit unexpectedly
+	nodeWatcherDone := make(chan struct{})
```
fernandol-nvidia: prob don't need dedicated done chans. See below

src/operator/resource_listener.go
```diff
@@ -0,0 +1,526 @@
...
+			return
+		case <-progressTicker.C:
+			// Report progress periodically even when idle
+			progressWriter := rl.GetProgressWriter()
+			if progressWriter != nil {
+				if err := progressWriter.ReportProgress(); err != nil {
+					log.Printf("Warning: failed to report progress: %v", err)
+				}
+			}
+		case msg := <-nodeChan:
```
fernandol-nvidia: msg, ok := <-nodeChan

if !ok {
  // nodeChan has closed
}

let the nodeChan and usageChan tell us when they are done instead of coordinating states from the parent.

src/operator/utils/node_usage_aggregator.go
ethany-nv: These maps are all interconnected. Each time we get an event, multiple maps will get updated. When we stream events from the informer, the events are being received and handled in callbacks _sequentially_ already anyways

src/operator/utils/node_usage_aggregator.go
ethany-nv: https://github.com/NVIDIA/OSMO/pull/281#discussion_r2748279427

fernandol-nvidia (approved):

PR #307, opened 2026-01-27
Remove action permissions from pool config

RyaliNvidia:
Issue #291

## Checklist

### DIFF
--- a/src/service/core/workflow/workflow_service.py
+++ b/src/service/core/workflow/workflow_service.py
@@ -495,10 +495,7 @@ def restart_workflow(pool_name: str,
 def cancel_workflow(name: str,
                     message: str | None = None,
                     force: bool = False,
-                    user_header: Optional[str] =
-                        fastapi.Header(alias=login.OSMO_USER_HEADER, default=None),
-                    roles_header: Optional[str] =
-                        fastapi.Header(alias=login.OSMO_USER_ROLES, default=None)) -> \
+                    user: str = fastapi.Depends(connectors.parse_username)) -> \
         objects.CancelResponse:
     """ Cancels the workflow. """
 


src/service/core/workflow/workflow_service.py
```diff
@@ -1092,34 +1035,20 @@ def action_request_helper(action_type: ActionType, payload: Dict[str, Any], name
 
 
 @router.post('/api/workflow/{name}/exec/group/{group_name}')
-def exec_into_group(name: str, group_name: str, entry_command: str,
-                    user_header: Optional[str] =
-                        fastapi.Header(alias=login.OSMO_USER_HEADER, default=None),
-                    roles_header: Optional[str] =
-                        fastapi.Header(alias=login.OSMO_USER_ROLES, default=None)) -> \
+def exec_into_group(name: str, group_name: str, entry_command: str) -> \
```
xutongNV, line 1038: this will be a breaking feature right?

src/service/core/workflow/workflow_service.py
```diff
@@ -1162,35 +1084,21 @@ def port_forward_task(name: str, task_name: str,
 
 
 @router.post('/api/workflow/{name}/webserver/{task_name}')
-def port_forward_webserver(name: str, task_name: str, task_port: int,
-                          user_header: Optional[str] =
-                            fastapi.Header(alias=login.OSMO_USER_HEADER, default=None),
-                          roles_header: Optional[str] =
-                            fastapi.Header(alias=login.OSMO_USER_ROLES, default=None)) -> \
+def port_forward_webserver(name: str, task_name: str, task_port: int) -> \
```
xutongNV, line 1087: please check if UI always pass these two args

src/utils/connectors/postgres.py
```diff
@@ -3084,7 +3067,6 @@ class PoolBase(pydantic.BaseModel):
     max_exec_timeout: str = ''
     max_queue_timeout: str = ''
     default_exit_actions: Dict[str, str] = {}
-    action_permissions: ActionPermissions = ActionPermissions()
     resources: PoolResources = PoolResources()
 
 class PoolMinimal(PoolBase):
```
xutongNV, line 3072: do you need `extra=pydantic.Extra.ignore` ?

src/service/core/workflow/workflow_service.py
RyaliNvidia: No. Removing action_permissions and removing these fields is not a breaking change

src/service/core/workflow/workflow_service.py
RyaliNvidia: Seems like it does. My change doesn't alter the behavior here with the arguments

src/utils/connectors/postgres.py
RyaliNvidia: I don't think we need to because anything which inherits this object already has it specified to ignore.

PR #338, opened 2026-01-30
#290 - Add attribute fetching for workflow pool matching

RyaliNvidia:
## Description

Issue #290

## Checklist

### DIFF
--- a/src/utils/roles/action_registry.go
+++ b/src/utils/roles/action_registry.go
@@ -197,6 +197,7 @@ var ActionRegistry = map[string][]EndpointPattern{
 	},
 	ActionWorkflowPortForward: {
 		{Path: "/api/workflow/*/portforward/*", Methods: []string{"*"}},
+		{Path: "/api/workflow/*/webserver/*", Methods: []string{"*"}},
 		{Path: "/api/router/portforward/*/client/*", Methods: []string{"*"}},
 		{Path: "/api/router/webserver/*", Methods: []string{"GET"}},
 	},


src/service/authz_sidecar/server/authz_server.go
```diff
@@ -42,36 +44,60 @@ const (
...
 	}
 }
 
+// MigrateRoles converts all legacy roles to semantic format and updates the database.
+// This should be called at startup to ensure all roles are in semantic format.
+func (s *AuthzServer) MigrateRoles(ctx context.Context) error {
+	// Get all role names from the database
+	allRoleNames, err := roles.GetAllRoleNames(ctx, s.pgClient)
+	if err != nil {
+		return fmt.Errorf("failed to get all role names: %w", err)
```
xutongNV, line 70: nit: should you align all fmt.Errorf to s.logger.Error?

src/utils/roles/action_registry.go
```diff
@@ -192,17 +193,28 @@ var ActionRegistry = map[string][]EndpointPattern{
 	},
 	ActionWorkflowExec: {
 		{Path: "/api/workflow/*/exec", Methods: []string{"POST", "WEBSOCKET"}},
+		{Path: "/api/router/exec/*/client/*", Methods: []string{"*"}},
 	},
 	ActionWorkflowPortForward: {
 		{Path: "/api/workflow/*/portforward/*", Methods: []string{"*"}},
```
xutongNV, line 199: `/api/workflow/*/webserver/*`

RyaliNvidia: Thanks for catching that

src/service/authz_sidecar/server/authz_server.go
RyaliNvidia: fmt is used to format the error string while logger is just printing to the console. This error is eventually pushed upstream and printed

xutongNV (approved):

src/utils/roles/roles.go
```diff
@@ -207,3 +207,55 @@ func GetAllRoleNames(ctx context.Context, client *postgres.PostgresClient) ([]st
...
+		policyBytes, err := json.Marshal(policy)
+		if err != nil {
+			return fmt.Errorf("failed to marshal policy for role %s: %w", role.Name, err)
+		}
+		policiesJSON[i] = policyBytes
+	}
+
+	// Update the role's policies in the database
+	// PostgreSQL expects JSONB[] which we construct from individual JSON values
+	query := `UPDATE roles SET policies = $1::jsonb[] WHERE name = $2`
```
ethany-nv, line 244: Should we consider having a separate table that has each policy as a row, instead of shoving all the policies into a JSONB?

src/utils/roles/action_registry.go
```diff
@@ -614,68 +613,145 @@ func MatchMethod(requestMethod string, allowedMethods []string) bool {
...
+		}
+		// Pool-scoped resources - workflow/task are scoped to pool
+		return extractWorkflowPoolResource(ctx, parts, actionName, pgClient)
 
-	// Internal resources - scoped to backend/workflow
 	case ResourceTypeInternal:
-		if len(parts) >= 3 {
-			return "backend/" + parts[2]
+		// Backend-scoped resources - internal actions
+		if actionName == "Operator" {
```
ethany-nv, line 662: consider using enum instead of string

src/utils/roles/action_registry.go
```diff
@@ -614,68 +613,145 @@ func MatchMethod(requestMethod string, allowedMethods []string) bool {
...
+		}
+	}
+
+	// If no workflow_id found or no postgres client, return wildcard
+	if workflowID == "" || pgClient == nil {
+		return string(ResourceTypePool) + "/*"
+	}
+
+	// Look up pool from workflow_id
+	poolName, err := GetPoolForWorkflow(ctx, pgClient, workflowID)
```
ethany-nv, line 738: Every time we need to evaluate the permissions for an API call associated with a workflow, we query the database? Can we minimize the calls to the DB? could we potentially cache this information?

RyaliNvidia: I'm currently caching the roles information so this is only making 1 db call per api request which is the same as what we are doing today. If this becomes more of an issue, we can further cache in the future, but I'm certain that this isn't a bottleneck right now

src/utils/roles/roles.go
RyaliNvidia: I don't think there is much to gain with splitting it up at the policy level because we don't fetch an individual policy of a role

src/utils/roles/action_registry.go
RyaliNvidia: Right now, this is only used here and at the definition so I felt like it was overkill to enum every single part of the action. If it turns out that I need to check for this field elsewhere, I'll make it a enum

src/utils/roles/action_registry.go
ethany-nv: Does caching the role information save us from making an API call to fetch pool? I can potentially see a user making many API calls to get information for just one workflow

ethany-nv (approved):

PR #364, opened 2026-02-04
setup listener for kind deployment

xutongNV:
## Description

Issue #147 

## Checklist

### DIFF
--- a/deployments/charts/backend-operator/README.md
+++ b/deployments/charts/backend-operator/README.md
@@ -37,6 +37,7 @@ This Helm chart deploys the OSMO Backend-Operator for managing compute backend r
 | `global.backendNamespace` | Backend namespace | `osmo-namespace` |
 | `global.backendTestNamespace` | Namespace for backend cluster validation tests | `null` |
 | `global.serviceUrl` | Service URL | `""` (empty, must be configured) |
+| `global.serviceGrpcPort` | Port for gRPC connections. Ignore if empty | `""` (empty) |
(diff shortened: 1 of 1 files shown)


deployments/charts/backend-operator/templates/backend-listener.yaml
```diff
@@ -80,7 +80,11 @@ spec:
         - listener
         args:
         - --serviceURL
+        {{- if .Values.global.serviceGrpcPort }}
```
ethany-nv, line 83: Is this not always true? is the else for backwards compatibility, using our python listener?

xutongNV: no. It is only need for local kind deployment. On aws we use the same port fot http and grpc

run/minimal/backend_operator_values.yaml
```diff
@@ -19,6 +19,8 @@ global:
   osmoImageTag: latest
   imagePullSecret: imagepullsecret
   serviceUrl: http://ingress-nginx-controller.ingress-nginx.svc.cluster.local
+  # serviceGrpcPort for backend-listener (gRPC to operator-service on TCP)
+  serviceGrpcPort: 50051
```
vvnpn-nv, line 23: can this be any port or does the service only support this specific port

RyaliNvidia (approved):

run/minimal/backend_operator_values.yaml
xutongNV: it can be any

Issue #354, opened 2026-02-03
Topology Aware Scheduling Implementation


ecolternv:
Complete the implementation for topology aware scheduling with KAI scheduler

PR #368, opened 2026-02-04
#354 - Implement Topology Aware Scheduling

ecolternv:
## Description
Provide support for topology aware scheduling
- Modify workflow spec to allow for definining topology rules
- Modify BackendSynchronizeQueues to support updating arbitrary resources, like topology
- Add support for topology keys in the pool spec
- Add logic to implement topology keys in the pod group for submitted workflows

Issue #354 

## Checklist
- [X] New or existing tests cover these changes. (Tracked in a separate issue)
- [X] The documentation is up to date with these changes. (Tracked in a separate issue)

### DIFF
--- a/src/utils/job/topology.py
+++ b/src/utils/job/topology.py
@@ -266,10 +266,11 @@ class PodGroupTopologyBuilder:
         # Walk down while there's only one child (shared by all tasks)
         while len(current.children) == 1:
             child = current.children[0]
-            top_level_constraint = {
-                ('requiredTopologyLevel' if child.required
-                 else 'preferredTopologyLevel'): child.label
-            }
+            if child.label is not None:
+                top_level_constraint = {
+                    ('requiredTopologyLevel' if child.required
+                     else 'preferredTopologyLevel'): child.label
+                }
             current = child
 
         return top_level_constraint, current
@@ -292,6 +293,9 @@ class PodGroupTopologyBuilder:
 
             # Leaf node - create subgroup for tasks
             if not node.children and node.tasks:
+                # Skip if node has no label or name (shouldn't happen for non-root nodes)
+                if node.label is None or name is None:
+                    return None
                 subgroup = {
                     'name': name,
                     'minMember': len(node.tasks),
@@ -313,6 +317,9 @@ class PodGroupTopologyBuilder:
 
             # Non-leaf - create parent subgroup if multiple children or has parent
             if len(node.children) > 1 or parent_name is not None:
+                # Skip if node has no label or name (shouldn't happen for non-root nodes)
+                if node.label is None or name is None:
+                    return None
                 subgroup = {
                     'name': name,
                     'topologyConstraint': {
--- a/src/utils/job/kb_objects.py
+++ b/src/utils/job/kb_objects.py
@@ -424,7 +424,7 @@ class KaiK8sObjectFactory(K8sObjectFactory):
         }
         pod_group_labels.update(labels)
 
-        pod_group_spec = {
+        pod_group_spec: Dict[str, Any] = {
             'queue': queue,
             'priorityClassName': priority_class,
         }


src/utils/job/backend_jobs.py
```diff
@@ -471,8 +477,10 @@ def execute(self, context: BackendJobExecutionContext,
         return JobResult()
 
 
+
```
RyaliNvidia: NIT: remove extra line

src/utils/job/topology.py
```diff
@@ -0,0 +1,134 @@
...
+        Algorithm:
+        1. Collect all topology requirements from all tasks
+        2. For each task, create TaskTopologyConstraint objects
+        3. Build a GroupTopologyConstraints mapping task names to their constraints
+
+        Args:
+            tasks: List of tasks in the group
+            pods: List of pod specs (in same order as tasks)
+            pool_name: Name of the pool
+            namespace: K8s namespace for the backend
```
RyaliNvidia: Will this field be used in the future?

src/utils/job/task.py
```diff
@@ -1965,8 +1965,28 @@ def get_kb_specs(
             progress_iter_freq,
 
         )
-        group_objects = k8s_factory.create_group_k8s_resources(group_uid, pods, labels, pool,
-                                                               priority)
+
+        # Build topology constraints if pool has topology enabled
+        # pylint: disable=import-outside-toplevel
+        from src.utils.job import topology as topology_module
```
RyaliNvidia: Do we want to just import this here or at the top level?

src/utils/job/topology.py
```diff
@@ -0,0 +1,134 @@
...
+
+class TopologyConstraintBuilder:
+    """Builds topology constraints for PodGroups from workflow specs"""
+
+    def __init__(self, pool: connectors.Pool):
+        """
+        Args:
+            pool: Pool configuration containing topology_keys
+        """
+        self.pool = pool
```
RyaliNvidia: I don't think you need to store the pool object in this class. When checking if the pool has no topology keys, you can simply check if key_to_label is empty

src/utils/job/task.py
```diff
@@ -1965,8 +1965,28 @@ def get_kb_specs(
...
+        pool_obj = connectors.Pool.fetch_from_db(self.database, pool)
+
+        builder = topology_module.TopologyConstraintBuilder(pool_obj)
+
+        # Validate all tasks' topology requirements
+        for task_obj in self.spec.tasks:
```
RyaliNvidia: You could run the validation inside the build constraints where if the task has a invalid key, you throw the ValueError

src/utils/job/topology.py
```diff
@@ -0,0 +1,134 @@
...
+
+        return GroupTopologyConstraints(
+            task_topology_constraints=task_topology_constraints
+        )
+
+    def validate_topology_requirements(self, resource_spec: connectors.ResourceSpec):
+        """Validates that topology requirements reference valid pool keys"""
+        for req in resource_spec.topology:
+            if req.key not in self.key_to_label:
+                raise ValueError(
```
RyaliNvidia: We should probably run osmo_errors.OSMOUserError instead of ValueError. Or like a validation error. One thing is that this function should probably be ran during the validation step of the workflow submission instead of during object instanciation

src/utils/job/task.py
```diff
@@ -1965,8 +1965,38 @@ def get_kb_specs(
...
+            topology_module.TopologyKey(key=tk.key, label=tk.label)
+            for tk in pool_obj.topology_keys
+        ] if pool_obj.topology_keys else []
+
+        # Convert tasks to TaskTopology with topology requirements
+        task_infos = []
+        for task_obj in self.spec.tasks:
+            topology_reqs = []
+            if task_obj.resources.topology:
+                for req in task_obj.resources.topology:
```
RyaliNvidia: Something feels wrong here. We don't need to look at the task object resource information. We just need to check the topology requirements once from the top level workflow resources spec object

src/utils/job/task.py
```diff
@@ -1965,8 +1965,38 @@ def get_kb_specs(
...
+                        req.requirementType == connectors.TopologyRequirementType.REQUIRED
+                    )
+                    topology_reqs.append(topology_module.TopologyRequirement(
+                        key=req.key,
+                        group=req.group,
+                        required=is_required
+                    ))
+            task_infos.append(topology_module.TaskTopology(
+                name=task_obj.name,
+                topology_requirements=topology_reqs
```
RyaliNvidia: requirements should be the same across all the tasks. Same idea from my previous comment

src/utils/job/topology.py
ecolternv: Removed

src/utils/job/task.py
ecolternv: Fixed. Added a rule to my CLAUDE.md to prevent this from happening again

src/utils/job/task.py
```diff
@@ -1965,8 +1965,38 @@ def get_kb_specs(
             progress_iter_freq,
 
         )
-        group_objects = k8s_factory.create_group_k8s_resources(group_uid, pods, labels, pool,
-                                                               priority)
+
+        # Build topology constraints if pool has topology enabled
+        pool_obj = connectors.Pool.fetch_from_db(self.database, pool)
```
ethany-nv: You can add a check here to see if this task uses the topology feature. If it doesn't, then we don't make DB call.

src/utils/job/topology.py
```diff
@@ -0,0 +1,345 @@
...
+    """
+    # Build key mappings
+    available_keys = {topology_key.key for topology_key in topology_keys}
+
+    # Collect unique key sets from tasks
+    key_sets = set()
+    for task in tasks:
+        if task.topology_requirements:
+            keys = tuple(sorted(req.key for req in task.topology_requirements))
+            key_sets.add(keys)
```
ethany-nv, line 97: Can we prematurely exit once key_sets length is greater than 1?

src/utils/job/task.py
RyaliNvidia: I guess in my head, there is x number of topologies and n number of tasks where 99.9% of the time x <= n so we shouldn't need to create the mapping where every duplicated topology is saved. It might be more work to make this slight optimization, but just something that felt wrong when I was reading through this

ecolternv: Yeah, I looked into this and tried a few different methods, but ultimately its too difficult to plumb that information down without a larger refactor. By the time we have split up the workflow spec into tasks, we no longer have a big deduplicated global list of resources saved like we do in the workflow spec. We only have resource objects attached to each task, where there is some duplication.

I think we can definitely simplify how we handle task/group/workflow objects, but that will require a larger refactor and is outside the scope of this change

src/utils/job/task.py
ecolternv: Same as above, I think this will trigger a much large refactor

src/utils/job/task.py
ecolternv: Done

src/utils/job/topology.py
ecolternv: I'd prefer to keep it this way, so that we can give a more complete error message (We list all unique key sets instead of just the first two unique ones we find)

RyaliNvidia (dismissed):

RyaliNvidia (approved):

PR #373, opened 2026-02-04
#294 #295 - Add user table and user-role mapping

RyaliNvidia:
## Description
This PR
- Adds new users table which store the user detailed information
- Adds two new tables for mapping user to roles and access tokens to roles
- Removes service access tokens since admins can now create a new user and then create a PAT for that user
- Adds a new field to roles to determine how the role is synced from the IdP if one is connected

Issue #294 #295

## Checklist

### DIFF
--- a/src/service/core/auth/auth_service.py
+++ b/src/service/core/auth/auth_service.py
@@ -662,8 +662,8 @@ def assign_role_to_user(
     try:
         result = postgres.execute_fetch_command(
             insert_cmd, (user_id, request.role_name, assigned_by, now), True)
-    except osmo_errors.OSMODatabaseError:
-        raise osmo_errors.OSMOUserError(f'User {user_id} not found')
+    except osmo_errors.OSMODatabaseError as err:
+        raise osmo_errors.OSMOUserError(f'User {user_id} not found') from err
 
     row = result[0]
     return objects.UserRoleAssignment(
--- a/src/utils/connectors/postgres.py
+++ b/src/utils/connectors/postgres.py
@@ -4354,7 +4354,8 @@ class Role(role.Role):
 
         # Determine sync parameters:
         # - external_roles_provided: True if self.external_roles is not None
-        # - external_roles_list: the list to use (empty if None, to be replaced by default for new roles)
+        # - external_roles_list: the list to use
+        #   (empty if None, to be replaced by default for new roles)
         external_roles_provided = self.external_roles is not None
         external_roles_list = self.external_roles if external_roles_provided else []
 
--- a/src/service/core/auth/tests/BUILD
+++ b/src/service/core/auth/tests/BUILD
@@ -1,0 +1,33 @@
+"""
+SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+SPDX-License-Identifier: Apache-2.0
+"""
+
+load("//bzl:py.bzl", "osmo_py_test")
+
+osmo_py_test(
+    name = "test_auth_service",
+    srcs = ["test_auth_service.py"],
+    deps = [
+        "//src/service/core/auth:auth",
+        "//src/service/core/tests:fixture",
+        "//src/tests/common",
+        "//src/utils/connectors",
+    ],
+    size = "large",
+    tags = ["requires-network"],
+    visibility = ["//visibility:public"],
+)
--- a/src/service/core/auth/tests/__init__.py
+++ b/src/service/core/auth/tests/__init__.py
@@ -1,0 +1,17 @@
+"""
+SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+SPDX-License-Identifier: Apache-2.0
+"""
--- a/src/service/core/auth/tests/test_auth_service.py
+++ b/src/service/core/auth/tests/test_auth_service.py
@@ -1,0 +1,745 @@
+"""
+SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+SPDX-License-Identifier: Apache-2.0
+"""
+
+from typing import Any, Dict, List, Optional
+
+from src.service.core.auth import objects
+from src.service.core.tests import fixture
+from src.utils import connectors
+from src.tests.common import runner
+
+
+class AuthServiceTestCase(fixture.ServiceTestFixture):
+    """Integration tests for auth service user and role management."""
+
+    TEST_USER = 'test@nvidia.com'
+    TEST_ADMIN = 'admin@nvidia.com'
+
+    def setUp(self):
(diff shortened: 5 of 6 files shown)


projects/PROJ-148-auth-rework/PROJ-148-user-management.md
```diff
@@ -0,0 +1,1134 @@
...
+
+```sql
+CREATE TABLE users (
+    -- Primary identifier (IDP username or service account name)
+    id TEXT PRIMARY KEY,
+
+    -- Metadata
+    created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+    created_by TEXT,  -- Username of who created this record
+    last_seen_at TIMESTAMP WITH TIME ZONE  -- Last activity timestamp (auto-updated on profile access)
```
ecolternv: Is this updated everytime the user calls any API? This is a lot of write traffic to the DB for something that isnt too important.

projects/PROJ-148-auth-rework/PROJ-148-user-management.md
```diff
@@ -0,0 +1,1134 @@
...
+
+-- Step 2: Populate users from existing profile and access_token data
+INSERT INTO users (id, created_at, created_by)
+SELECT DISTINCT user_name, NOW(), 'migration'
+FROM profile WHERE user_name NOT IN (SELECT id FROM users)
+ON CONFLICT (id) DO NOTHING;
+
+INSERT INTO users (id, created_at, created_by)
+SELECT DISTINCT user_name, NOW(), 'migration'
+FROM access_token WHERE access_type = 'USER'
```
ecolternv: Shouldn't this be access_type = 'SERVICE'? We want to create a "service user" for all service access tokens

src/service/core/auth/auth_service.py
```diff
@@ -119,12 +122,15 @@ def get_jwt_token_from_access_token(access_token: str):
     if token.expires_at.date() <= datetime.datetime.utcnow().date():
         raise osmo_errors.OSMOUserError('Access Token has expired')
 
+    # Get roles from pat_roles table
+    roles = objects.AccessToken.get_roles_for_token(postgres, token.user_name, token.token_name)
```
ecolternv, line 126: Is it enough to get roles from the token and trust those? Shouldn't we also double check the user roles and make sure the user has all of those roles as well?

ecolternv: If we cant do that right now, then we should have some documentation about this, otherwise and admin might remove a role from a user and be surprised they can still use that role.

src/service/core/auth/objects.py
```diff
@@ -135,26 +101,176 @@ def insert_into_db(cls, database: connectors.PostgresConnector, user_name: str,
...
+
         try:
-            database.execute_commit_command(
-                insert_cmd,
-                (user_name, token_name, auth.hash_access_token(access_token), expires_at,
-                 description, access_type.value, roles))
-        except osmo_errors.OSMODatabaseError as e:
-            raise osmo_errors.OSMOUserError(f'Token name {token_name} already exists.') from e
+            database.execute_commit_command(insert_cmd, args)
+        except Exception as e:
```
ecolternv: Can we catch something more specific than generic Exception? Is there some top level postgres exception we can catch?

src/service/core/auth/objects.py
```diff
@@ -135,26 +101,171 @@ def insert_into_db(cls, database: connectors.PostgresConnector, user_name: str,
...
+            error_str = str(e).lower()
+            if 'already exists' in error_str or 'duplicate key' in error_str:
```
fernandol-nvidia, line 163: Is `from psycopg2.errors import UniqueViolation` available to us?

src/service/core/auth/objects.py
```diff
@@ -135,26 +101,171 @@ def insert_into_db(cls, database: connectors.PostgresConnector, user_name: str,
...
+                FROM token_insert ti
+                INNER JOIN user_roles ur ON ur.user_id = ti.user_name
+                WHERE ur.role_name = ANY(%s::text[])
+                RETURNING role_name
+            )
+            SELECT
+                CASE
+                    WHEN (SELECT COUNT(*) FROM role_insert) = %s
+                    THEN 'ok'
+                    ELSE (SELECT (1/0)::text)
```
fernandol-nvidia: consider a more explicit sentinel value (e.g. "error_missing_roles") and check for that in error handling (instead of relying on Postgres to always return msg with "division by zero"

src/utils/connectors/postgres.py
```diff
@@ -1093,35 +1101,112 @@ def _init_tables(self):
         '''
         self.execute_commit_command(create_cmd, ())
 
-        # Creates table for access token keys.
+        # Creates table for users (IDP users and service accounts)
+        create_cmd = '''
+            CREATE TABLE IF NOT EXISTS users (
+                id TEXT PRIMARY KEY,
```
fernandol-nvidia, line 1107: will we need to handle multi-IDP at some point? If so, userId (e.g. a@123.com) alone might not be enough. And we might need a IDP column as well. If not, then I think this is fine.

src/service/core/auth/auth_service.py
```diff
@@ -141,93 +147,716 @@ def get_access_token_info(access_token: str) -> objects.AccessToken:
...
+        request: AssignRoleRequest with role_name
+        assigned_by: Authenticated user making the request
+
+    Returns:
+        UserRoleAssignment with assignment details
+    """
+    postgres = connectors.PostgresConnector.get_instance()
+
+    # Validate user and role exist
+    _validate_user_exists(postgres, user_id)
```
ecolternv: NIT, but is a lot of this validation technically redundant? We have foreign constraints in teh table taht should catch username/role not existing when we insert user_roles

src/utils/connectors/postgres.py
```diff
@@ -1522,6 +1607,110 @@ def fetch_user_names(self, user_names: List[str]) -> List[str]:
...
+    upsert_cmd = '''
+        INSERT INTO users (id, created_at, created_by, last_seen_at)
+        VALUES (%s, NOW(), %s, NOW())
+        ON CONFLICT (id) DO UPDATE SET
+            last_seen_at = NOW();
+    '''
+    database.execute_commit_command(upsert_cmd, (user_name, user_name))
+
+
+def sync_user_roles(database: PostgresConnector, user_name: str, external_roles: List[str]):
```
fernandol-nvidia, line 1623: Can this whole thing be a single DB transaction so it is atomic?

Right now, each `database.execute_commit_command` is its own transaction. @ethany-nv (for viz since we recently added connection pooling and updated a few postgres commands.. We might want to expose https://github.com/NVIDIA/OSMO/blob/f06d22c8a5e2cb65a3e677bc6885b823731c8d3a/src/utils/connectors/postgres.py#L302 so anyone can compose their own transaction in a conn.

src/utils/connectors/postgres.py
```diff
@@ -4079,31 +4340,79 @@ def delete_from_db(cls, database: PostgresConnector, name: str):
         database.execute_commit_command(delete_cmd, (name,))
 
     def insert_into_db(self, database: PostgresConnector, force: bool = False):
```
fernandol-nvidia, line 4341: Similarly, there are multiple transactions happening here.

projects/PROJ-148-auth-rework/PROJ-148-user-management.md
RyaliNvidia: Yeah. I can get rid of this then

projects/PROJ-148-auth-rework/PROJ-148-user-management.md
RyaliNvidia: This part of the doc can be removed. It is stale. I have an updated migration sql file which is correct

fernandol-nvidia (commented):
Design and logic lgtm overall. I think the biggest thing is atomic transaction is not applied consistently across the implementation (also not unique to this PR but it is something the codebase has a long history of..). But if we have an opportunity to do it here (since its a new service), that would be ideal.

src/service/core/auth/auth_service.py
RyaliNvidia: When an access_token is created, it verifies that all roles mapped are part of the user and when the user has a role deleted, it is an atomic operation with the deletion of the user role mapping. I was considering creating a uuid for the user-role mapping table which will then be mapped to the access token table so we could have a on cascade delete. Do you think this is a good idea?

src/service/core/auth/objects.py
RyaliNvidia: Yep. Updating

src/utils/connectors/postgres.py
RyaliNvidia: I'm saying no to multi-IdP right now, but also we have no way of figuring out the IdP from the request right now so I'm not sure if there is a good way of adding it to the db

src/utils/connectors/postgres.py
RyaliNvidia: updated

src/utils/connectors/postgres.py
RyaliNvidia: updated

src/service/core/auth/objects.py
RyaliNvidia: That is currently being masked inside of execute_commit_command

src/service/core/auth/objects.py
RyaliNvidia: I'm considering making a uuid for every entry in the user_role mapping table where the pat_roles table will have a foreign key reference so if the entry is removed in the user_role, it would error here. I think that might be a more elegant solution

RyaliNvidia: Yeah. I'll do this. I think it simplifies some code

src/service/core/auth/auth_service.py
RyaliNvidia: I'm going ahead with the uuid field because I think it makes the relationship make more sense

src/service/core/auth/auth_service.py
RyaliNvidia: Updated some of the cases

PR #380, opened 2026-02-05
Add more message types

xutongNV:
- add pod monitor
- add pod condition

## Description

Issue #147 

## Checklist

### DIFF
--- a/src/service/agent/helpers.py
+++ b/src/service/agent/helpers.py
@@ -433,8 +433,10 @@ def keep_pod_conditions(message: backend_messages.ConditionMessage) -> bool:
     return True
 
 
-def send_pod_conditions(message: backend_messages.PodConditionsBody,
-                        max_event_log_lines: int):
+def send_pod_conditions(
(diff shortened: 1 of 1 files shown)


src/service/agent/message_worker.py
```diff
@@ -196,7 +196,11 @@ def process_message(self, message_id: str, message_json: str, backend_name: str)
             message_body = backend_messages.MessageOptions(**message_options)
 
             if message_body.update_pod:
+                if message_body.update_pod.status == task.TaskGroupStatus.INITIALIZING:
+                    helpers.create_monitor_job(message_body.update_pod)
```
ethany-nv, line 200: Do we need monitor_job? After looking into it, maybe it's not useful. I think originally it was meant to address 'CreateContainerConfigError', but that logic should be defined when evaluating a pod's waiting_reason in workflow listener.

xutongNV: It is no harm to have this so we dont missing unexpected cases

ethany-nv (approved):

RyaliNvidia (approved):

PR #385, opened 2026-02-05
Detect non-aws S3 compatible endpoints when validating data_auth

fernandol-nvidia:
## Description
- To use non-AWS S3 compatible storage, we set `AWS_ENDPOINT_URL_S3` vs `AWS_ENDPOINT_URL`
- However, our s3 data-auth will always attempt to validate against AWS S3
- We need to detect when non-AWS S3 compatible storage is used and perform our validation accordingly

Issue - None

## Tested
- We currently use localstack (S3 compatible) with `OSMO_SKIP_DATA_AUTH` enabled
- We can remove `OSMO_SKIP_DATA_AUTH` completely and test that localstack still works in quickstart.

## Checklist

### DIFF
--- a/src/lib/data/storage/backends/tests/test_backends.py
+++ b/src/lib/data/storage/backends/tests/test_backends.py
@@ -294,27 +294,27 @@ class ExtractAccountKeyFromConnectionStringTest(unittest.TestCase):
         self.assertIn('AccountKey not found', str(context.exception))
 
 
-class IsS3CompatibleEndpointTest(unittest.TestCase):
-    """Tests for S3-compatible endpoint detection."""
+class IsNonAwsS3EndpointConfiguredTest(unittest.TestCase):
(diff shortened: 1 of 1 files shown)


src/lib/data/storage/backends/backends.py
```diff
@@ -52,6 +52,40 @@ def _skip_data_auth() -> bool:
     return os.getenv(OSMO_SKIP_DATA_AUTH, '0') == '1'
 
 
+def _is_s3_compatible_endpoint() -> bool:
```
RyaliNvidia: NIT: technically this doesn't check if it is s3 compatible. It just checks if the env is set and isn't AWS

fernandol-nvidia: Yup, renamed to _is_non_aws_s3_endpoint_configured. Thanks!

RyaliNvidia (approved):

PR #388, opened 2026-02-05
add envoy lua filter to refresh id_token

vvnpn-nv:
#### Problem: Token refresh missing IdToken causes "Jwt is missing" error

When using Microsoft Entra ID (Azure AD) as the OAuth2 provider, users experience a "Jwt is missing" error after approximately 5 minutes of inactivity. This issue occurs because:
* Microsoft's OAuth2 token refresh does not return id_token unless scope=openid is explicitly included in the refresh request
* Envoy's OAuth2 filter does not include scope=openid when performing token refresh
* The JWT authentication filter requires the IdToken cookie to validate reques

#### Solution: 
Added two Lua filters to the Envoy configuration for web-ui, service, and router charts:

1. Request-side filter (validate_idtoken)

  Runs before the OAuth2 filter to detect invalid auth state early:
  * Checks if BearerToken exists (proves user was previously authenticated)
  * Checks if IdToken is missing or malformed (not a valid JWT structure)
  * If detected, clears all cookies to force a full re-authentication

2. Response-side filter (is_missing_idtoken_on_refresh)

  Runs after the OAuth2 filter to catch bad refresh responses:
  * Detects when a refresh response contains BearerToken but no valid IdToken
  * Clears all auth cookies in the response to prevent persisting bad state

Issue - None

## Checklist

### DIFF
--- a/deployments/charts/router/templates/_envoy-config-helpers.tpl
+++ b/deployments/charts/router/templates/_envoy-config-helpers.tpl
@@ -159,6 +159,7 @@ listeners:
                   end
                 end
 
+        {{- if .Values.sidecars.envoy.oauth2Filter.forceReauthOnMissingIdToken }}
         - name: envoy.filters.http.lua.validate_idtoken
           typed_config:
             "@type": type.googleapis.com/envoy.extensions.filters.http.lua.v3.Lua
@@ -219,6 +220,7 @@ listeners:
                     end
                   end
                 end
+        {{- end }}
 
         - name: envoy.filters.http.lua.pre_oauth2
           typed_config:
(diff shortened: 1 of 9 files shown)


RyaliNvidia (approved):

deployments/charts/router/templates/_envoy-config-helpers.tpl
```diff
@@ -214,9 +248,54 @@ listeners:
...
                   local header = response_handle:headers():get("set-cookie")
+
                   if header ~= nil then
+                    -- Check if this is a token refresh response missing IdToken.
+                    -- unless scope=openid is explicitly included (which Envoy doesn't do).
+                    if is_missing_idtoken_on_refresh(header) then
+                      response_handle:logInfo("OAuth2 refresh missing IdToken - clearing session to force re-auth")
+
+                      -- Clear all auth cookies to force full re-authentication
+                      response_handle:headers():remove("set-cookie")
```
xutongNV, line 315: Will this remove not just the auth cookie, but also AWS stickness cookies?

vvnpn-nv: yes I think so, updated to only remove auth cookies

RyaliNvidia (approved):

PR #397, opened 2026-02-07
Refactor Node Condition Updates

xutongNV:
### Summary

Refactors node condition updates to use backend jobs instead of Redis queues and backend websocket messages. The backend worker now directly applies node condition rules and updates Kubernetes node labels, while the resource listener detects these changes via Kubernetes watch.

### Before: Redis Queue + Websocket Flow

```mermaid
graph LR
    A[Config Update] --> B[Redis Queue]
    B --> C[Agent Service]
    C --> D[Websocket Message]
    D --> E[Backend Listener]
    E --> F[Update Labels]
```

### After: Backend Job Flow

```mermaid
graph LR
    A[Config Update] --> B[Backend Job Queue]
    B --> C[Backend Worker]
    C --> D[Update Labels]
```

Issue #147 

## Checklist

### DIFF
--- a/src/service/core/config/helpers.py
+++ b/src/service/core/config/helpers.py
@@ -20,6 +20,7 @@ from collections import abc
 from datetime import datetime
 import json
 import logging
+import re
 from typing import Any, Dict, List
 
 import pydantic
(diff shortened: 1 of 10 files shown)


src/utils/job/backend_jobs.py
```diff
@@ -972,11 +974,151 @@ def execute(self, context: BackendJobExecutionContext,
...
+                    # Invalid regex should be ignored
+                    continue
+            # If at least one rule matched this condition type but none allowed the status,
+            # the node is not available.
+            if matched_any_rule and not allowed_by_any_rule:
+                return False
+
+        return not node.spec.unschedulable
+
+    def get_effective_rules(self, default_rules: Dict[str, str]) -> List[Tuple[str, str]]:
```
ethany-nv, line 1016: I noticed this function doesn't have underscore, but others do (assuming private function). Maybe we should standardize the helper functions for this class

src/service/core/config/helpers.py
```diff
@@ -243,10 +269,7 @@ def _update_backend_helper(
 
     # Check if the backend node_conditions has changed
     if send_update:
-        backend_action_request_helper(
-            payload=yaml.safe_load(configs['node_conditions']),
-            name=configs['name'],
-        )
+        backend_action_request_helper(name=configs['name'])
```
ethany-nv, line 272: You can consider passing the prefix and the node conditions here, so when backend_action_request_helper runs, we can make 1 less DB call to fetch

PR #398, opened 2026-02-07
Compress Backend Job when Service sends to Backend Worker

ethany-nv:
## Description
By using compression (such as zlib), we can send jobs using 1/5 to 1/10 of the memory footprint as before. For a workflow that has a lot of repetition, we can use 25x less memory. This can potentially unblock us for big workflow submissions before the re-arch.

For example, for a parallel workflow that has 1 server and 99 client tasks:

Before:
```
2026-02-07T01:55:16.745+00:00 worker [INFO] backend_worker: Memory usage of value: 984017 bytes (960.95 KB, 0.94 MB)
```

After compression:
```
2026-02-07T02:12:32.895+00:00 worker [INFO] backend_worker: Value: b'x\x9c\xec\xbd[s\xe3\xc8\x92\xe7\xf9U\xc6\xce\xf3v[\xdcp[\xb...'
2026-02-07T02:12:32.896+00:00 worker [INFO] backend_worker: Memory usage of value: 46400 bytes (45.31 KB, 0.04 MB)
```

Issue #411 

## Checklist

### DIFF
--- a/src/operator/backend_worker.py
+++ b/src/operator/backend_worker.py
@@ -23,9 +23,9 @@ import os
 import threading
 import time
 import traceback
-import zlib
 from typing import Dict, Optional
 from urllib.parse import urlparse
+import zlib
 
 import pydantic
 import websockets
--- a/src/service/agent/objects.py
+++ b/src/service/agent/objects.py
@@ -24,8 +24,8 @@ import json
 import logging
 import threading
 import time
-import zlib
 from typing import Callable, Optional, Dict, List, Set
+import zlib
 
 import fastapi
 import kombu  # type: ignore


fernandol-nvidia (approved):

src/service/agent/objects.py
```diff
@@ -253,7 +254,8 @@ async def handle_job(self, job_spec: Dict):
                     job=job,
                     start_time=time.time())
 
-            await self.websocket.send_text(job.json())
+            compressed = zlib.compress(job.json().encode('utf-8'))
```
xutongNV, line 257: how long compress takes, will this break progress liveness?

src/operator/backend_worker.py
```diff
@@ -23,6 +23,7 @@
 import threading
 import time
 import traceback
+import zlib
```
ecolternv: NIT: This should go after urlllib.parse for alphabetical ordering

src/service/agent/objects.py
```diff
@@ -24,6 +24,7 @@
 import logging
 import threading
 import time
+import zlib
```
ecolternv: Same here, it should go after `typing`

src/service/agent/objects.py
ethany-nv: For 100 tasks, around 88 ms

xutongNV (dismissed):

xutongNV (approved):

fernandol-nvidia (approved):

PR #408, opened 2026-02-09
#292 - Remove middleware service + add allow/deny effects

RyaliNvidia:
Issue #292

## Checklist

### DIFF
--- a/src/lib/utils/role.py
+++ b/src/lib/utils/role.py
@@ -69,19 +69,7 @@ class RolePolicy(pydantic.BaseModel):
     @classmethod
     def validate_actions(cls, value) -> List[str]:
         """Parse and validate actions from various input formats."""
-        result = []
-        for action in value:
-            if isinstance(action, str):
-                raw = action
-            elif isinstance(action, dict):
-                raw = action.get('action')
-                if raw is None:
-                    raise osmo_errors.OSMOUserError(
-                        'Invalid action dict: missing "action" key')
-            else:
-                raise osmo_errors.OSMOUserError(f'Invalid action type: {type(action)}')
-            result.append(validate_semantic_action(raw))
-        return result
+        return [validate_semantic_action(action) for action in value]
 
     def to_dict(self) -> Dict[str, Any]:
         """


cypres (dismissed):

src/lib/utils/role.py
```diff
@@ -17,113 +17,138 @@
...
+    def parse_actions_as_strings(cls, data: List[Dict]) -> List[Dict]:
+        """Parse the actions as strings for display purposes."""
         data_list = []
         for role in data:
-            role['policies'] = [{
-                'actions': [RoleAction.to_str(action['base'], action['path'], action['method'])
-                            for action in policy['actions']]
-            } for policy in role['policies']]
-            data_list.append(role)
+            role_copy = role.copy()
```
ethany-nv: consider using deepcopy()?

src/lib/utils/role.py
```diff
@@ -17,113 +17,138 @@
...
+            role_copy['policies'] = []
+            for policy in role['policies']:
+                effect = policy.get('effect', PolicyEffect.ALLOW)
```
ethany-nv: what if effect is not a valid PolicyEffect?

src/lib/utils/role.py
```diff
@@ -17,113 +17,138 @@
...
+                if isinstance(effect, PolicyEffect):
```
ethany-nv: Consider instantiating a role object for `role`, so that the code is not just fetching the key (like `role['policies']` and instantiating an actual object also does validation as well

src/utils/connectors/postgres.py
```diff
@@ -4075,74 +4035,53 @@ def insert_into_db(self, database: PostgresConnector, force: bool = False):
...
-                    role.RoleAction(base='http', path='/api/pool_quota', method='*'),
-                    role.RoleAction(base='http', path='/api/profile/*', method='*'),
-                    role.RoleAction(base='http', path='/api/users', method='*'),
-                    role.RoleAction(base='http', path='/api/tag', method='*'),
-                    role.RoleAction(base='http', path='/api/plugins/configs', method='*'),
-                    # Tailing slash is to exclude path /api/router/webserver/*/backend/*
-                    role.RoleAction(base='http', path='/api/router/webserver/*/', method='*'),
-                    role.RoleAction(base='http', path='/api/router/*/*/client/*', method='*'),
-                ]
+                    # Workflow actions
```
ethany-nv: Are the auth endpoints covered?

src/utils/roles/action_registry.go
```diff
@@ -1275,28 +1304,27 @@ func CheckPolicyAccess(
...
-		if result.Matched {
-			lastResult = result
-		}
 	}
 
-	// If we had a matched deny, return it
-	if lastResult.Matched {
-		return lastResult
+	// Second pass: first Allow wins
+	for _, role := range roles {
```
ethany-nv: Can you do it in one go, instead of two passes?

src/lib/utils/role.py
RyaliNvidia: This function is only used for showing the config history so if it isn't a valid PolicyEffect that is fine

src/lib/utils/role.py
RyaliNvidia: This function is used for showing the config history, so we shouldn't instantiate the object because the object schema could have changed and we don't want to accidentally cause a old invalid schema to break this api

src/utils/connectors/postgres.py
RyaliNvidia: Yeah. They are covered in the osmo-default role which every user inherits

src/utils/roles/action_registry.go
RyaliNvidia: Yep. Updating

src/lib/utils/role.py
RyaliNvidia: deepcopy isn't needed here because I'm only modifying the policies field, but it is fine because in the line right after it, I set it to a list and then add new entries to that list

cypres (approved):

PR #410, opened 2026-02-10
split node and resource

xutongNV:
## Description

Issue #147 

## Checklist

### DIFF
--- a/src/operator/node_usage_listener.go
+++ b/src/operator/node_usage_listener.go
@@ -33,41 +33,42 @@ import (
 	pb "go.corp.nvidia.com/osmo/proto/operator"
 )
 
-// ResourceListener manages the bidirectional gRPC stream for pod resource usage
-type ResourceListener struct {
+// NodeUsageListener manages the bidirectional gRPC stream for pod resource usage
+type NodeUsageListener struct {
 	*utils.BaseListener
 	args       utils.ListenerArgs
 	aggregator *utils.NodeUsageAggregator
 }
 
-// NewResourceListener creates a new resource listener instance
-func NewResourceListener(args utils.ListenerArgs) *ResourceListener {
-	return &ResourceListener{
-		BaseListener: utils.NewBaseListener(args, "last_progress_resource_listener", "resource"),
-		args:         args,
-		aggregator:   utils.NewNodeUsageAggregator(args.Namespace),
+// NewNodeUsageListener creates a new node usage listener instance
+func NewNodeUsageListener(args utils.ListenerArgs) *NodeUsageListener {
+	return &NodeUsageListener{
+		BaseListener: utils.NewBaseListener(
+			args, "last_progress_node_usage_listener", utils.StreamNameResource),
+		args:       args,
+		aggregator: utils.NewNodeUsageAggregator(args.Namespace),
 	}
 }
 
 // Run manages the bidirectional streaming lifecycle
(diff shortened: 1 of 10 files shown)


src/operator/node_usage_listener.go
```diff
@@ -21,7 +21,6 @@ import (
 	"fmt"
```
ethany-nv, line 21: Let's call this file node_usage_listener

src/operator/utils/base_listener.go
```diff
@@ -31,18 +31,25 @@ import (
...
+	ctx context.Context,
+	cancel context.CancelCauseFunc,
+	ch <-chan *pb.ListenerMessage,
+)
 
 // BaseListener contains common functionality for all listeners
 type BaseListener struct {
 	unackedMessages *UnackMessages
 	progressWriter  *progress_check.ProgressWriter
+	streamName      string
```
ethany-nv: should we make this type an enum?

ethany-nv (approved):

src/operator/utils/base_listener.go
```diff
@@ -139,8 +161,8 @@ func (bl *BaseListener) ReceiveAcks(ctx context.Context, cancel context.CancelCa
 	}
 }
 
-// WaitForCompletion waits for goroutines to finish
-func (bl *BaseListener) WaitForCompletion(ctx context.Context, streamCtx context.Context) error {
+// waitForCompletion waits for goroutines to finish
+func (bl *BaseListener) waitForCompletion(ctx context.Context, streamCtx context.Context) error {
 	// Wait for context cancellation (from parent or goroutines)
 	<-streamCtx.Done()
```
fernandol-nvidia, line 167: Not related to this PR, but it caught my eyes :P

We should wait for both streamCtx and ctx in case one of the finishes first so this method can participate in context cancellation, for example:

```
select {
case <-streamCtx.Done():
   ...
case <-ctx.Done():
  ...
}
```

src/operator/utils/base_listener.go
```diff
@@ -268,32 +277,44 @@ func (bl *BaseListener) Run(
...
 
-	// Launch goroutines for send and receive
-	bl.AddToWaitGroup(2)
+	// Launch three goroutines: receiveAcks, watch, sendMessages
+	bl.wg.Add(3)
 	go func() {
-		defer bl.WaitGroupDone()
+		defer bl.wg.Done()
 		defer func() {
 			if r := recover(); r != nil {
```
fernandol-nvidia, line 285: Sry, also not related to your PR, can you remind me why we are recovering from panic here? If something is truly a panic, I prefer to fail fast and fix the bugs...

src/operator/utils/base_listener.go
fernandol-nvidia, line 258: Since we are deriving a new context, we can just pass `streamCtx` into our functions (e.g. `watch`, `sendMessages`, etc). This also simplifies and addresses one of the comments above about needing to handle two ctx.

fernandol-nvidia (approved):
My comments are not blocking, but they should be addressed so our implementation can be more Golang idiomatic

src/operator/utils/base_listener.go
fernandol-nvidia: Refer to https://github.com/NVIDIA/OSMO/pull/410#discussion_r2796254712

PR #418, opened 2026-02-10
#148 - Add user mapping into OSMO

RyaliNvidia:
Issue #148

## Checklist

### DIFF
--- a/deployments/charts/router/templates/_envoy-config-helpers.tpl
+++ b/deployments/charts/router/templates/_envoy-config-helpers.tpl
@@ -104,6 +104,8 @@ listeners:
           internal_only_headers:
           - x-osmo-auth-skip
           - x-osmo-user
+          - x-osmo-token-name
+          - x-osmo-workflow-id
           virtual_hosts:
           - name: service
             domains: ["*"]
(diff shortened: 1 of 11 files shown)


src/cli/user.py
```diff
@@ -0,0 +1,247 @@
...
+                             help='Specify the output format type (Default text).')
+    list_parser.set_defaults(func=_list_users)
+
+    # Create user
+    create_parser = subparsers.add_parser(
+        'create',
+        help='Create a new user.',
+        description='Create a new user with optional roles.',
+        epilog='Ex. osmo user create myuser@example.com\n'
+               'Ex. osmo user create service-account --roles osmo-user osmo-ml-team',
```
ecolternv, line 63: NIT: not related to this change really, but we should consider changing the default roles to not have `osmo` in them. `osmo-user` -> `user` `osmo-admin` -> `admin`. The `osmo` in the name is redundant because we are alreayd in osmo, so of course hte role is related to OSMO.

I think this is a holdover from when we had `grafana-user` vs `osmo-user` in our SSO token that went to keycloak, but inside osmo we dont need the extra `osmo`

ecolternv (approved):

PR #421, opened 2026-02-11
Support Non-AWS S3 storage without environment variables

fernandol-nvidia:
Issue #383

## Checklist

### DIFF
--- a/run/run_command.py
+++ b/run/run_command.py
@@ -112,6 +112,7 @@ class Process:
             return False, self.stdout_file, self.stderr_file, elapsed_time
 
 
+
 def run_command_with_logging(
     cmd: List[str],
     description: Optional[str] = None,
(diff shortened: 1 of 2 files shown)


RyaliNvidia (approved):

deployments/charts/quick-start/templates/config-setup.yaml
```diff
@@ -67,6 +67,7 @@ spec:
                 "workflow_data": {
                   "credential": {
                     "endpoint": "'"${OBJECT_STORAGE_ENDPOINT}"'/workflows",
+                    "override_url": "'"${OBJECT_STORAGE_OVERRIDE_URL}"'",
```
vvnpn-nv, line 70: do these need the postfix `/workflows`?

fernandol-nvidia: Technically no. But if we don't, the top level bucket is going to be polluted with workflows/logs/apps. So this is a simple way to organize the objects our app writes.

PR #435, opened 2026-02-13
Backend Monitor Part 2 (backend side)

xutongNV:
## Description

Issue #147 

## Checklist

### DIFF
--- a/src/service/core/config/helpers.py
+++ b/src/service/core/config/helpers.py
@@ -237,10 +237,10 @@ def _update_backend_helper(
                         continue
             except json.JSONDecodeError as e:
                 raise osmo_errors.OSMOUserError(
-                    f"Invalid JSON in node_conditions: {e}") from e
+                    f'Invalid JSON in node_conditions: {e}') from e
             except (KeyError, TypeError) as e:
                 raise osmo_errors.OSMOUserError(
-                    f"Invalid node_conditions format: {e}") from e
+                    f'Invalid node_conditions format: {e}') from e
             send_update = True
         if key == 'tests' and value:
             # Check for duplicates in the original list


src/operator/utils/node_condition_rules.go
```diff
@@ -0,0 +1,96 @@
...
+		if matched, err := matchFromStart(pattern, condType); err == nil && matched {
+			return true
+		}
+	}
+	return false
+}
+
+// matchFromStart matches a pattern from the start of the string, similar to Python's re.match().
+// re.match() always matches from the beginning, regardless of whether the pattern has '^'.
+func matchFromStart(pattern, text string) (bool, error) {
```
ethany-nv: Do we need this? Can we just add '^' to the beginning of the rules to begin with?

src/operator/utils/node_condition_rules.go
```diff
@@ -0,0 +1,96 @@
...
+	for condType, statusRegex := range DefaultAvailableCondition {
+		if !hasMatchingPattern(effective, condType) {
+			// Escape the condition type and anchor it
+			escaped := regexp.QuoteMeta(condType)
+			effective["^"+escaped+"$"] = statusRegex
+		}
+	}
+
+	nc.mu.Lock()
+	nc.rules = effective
```
ethany-nv, line 73: u can just do defer on line 72 too, to stay consistent

src/operator/node_listener.go
```diff
@@ -119,7 +122,7 @@ func (nl *NodeListener) watchNodes(
 
 	nodeInformerFactory := informers.NewSharedInformerFactory(
 		clientset,
-		0, // No automatic resync
+		time.Duration(nl.args.ResyncPeriodSec)*time.Second,
```
ethany-nv, line 125: Do we need to resync?

xutongNV: Rsync helps refresh nodes when node_condition rule changes

cypres (approved):

src/operator/utils/node_condition_rules.go
```diff
@@ -0,0 +1,84 @@
...
+
+// DefaultAvailableCondition defines the default node condition rules for availability checking.
+var DefaultAvailableCondition = map[string]string{"Ready": "True"}
+
+// NodeConditionRules manages node condition rules in a thread-safe manner.
+// Rules format: map[regex]regex mapping condition.type regex to a status regex.
+// The status regex must be a combination of: True|False|Unknown (OR-ed).
+// The rules field stores effective rules (user rules + defaults).
+type NodeConditionRules struct {
+	rules map[string]string
```
ethany-nv, line 33: should we use sync.Map instead of map + RWMutex? For node conditions, it's mostly read operations, and rarely write operations

ethany-nv (commented):
comment

fernandol-nvidia (approved):

src/operator/utils/node_condition_rules.go
xutongNV: this is some analysis:
why using read-write lock instead of sync.map
- Entire map operations, not per-key
       GetRules() returns a copy of the entire map
       SetRules() replaces the entire map
       sync.Map is optimized for per-key access
- Read-heavy, write-rare pattern
       GetRules() is called on every node event (high frequency)
       SetRules() is called only when rules are updated (low frequency)
       sync.RWMutex allows multiple concurrent readers, which fits this pattern

PR #447, opened 2026-02-18
Client install location is determined by service

RyaliNvidia:
The service is no longer hosting the client so the messages need to be updated

Issue #None

## Checklist

### DIFF
--- a/src/service/core/service.py
+++ b/src/service/core/service.py
@@ -17,6 +17,7 @@ SPDX-License-Identifier: Apache-2.0
 """
 
 import base64
+import datetime
 import logging
 from pathlib import Path
 import sys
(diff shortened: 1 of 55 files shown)


fernandol-nvidia (approved):

src/service/core/service.py
```diff
@@ -280,6 +291,21 @@ def set_default_service_url(postgres: connectors.PostgresConnector):
             postgres.config.service_hostname)
 
 
+def set_client_install_url(postgres: connectors.PostgresConnector,
+                           config: objects.WorkflowServiceConfig):
+    curr_service_configs = postgres.get_service_configs()
+    if curr_service_configs.cli_config.client_install_url != config.client_install_url:
```
ethany-nv, line 298: What if `config.client_install_url` is empty?

RyaliNvidia: https://github.com/NVIDIA/OSMO/pull/447/changes#diff-9de630e332c47cce6bb3d9515be435e8bbdeba82bdc6900e5da084a18aed9dbfR69
The message is different

Issue #356, opened 2026-02-03
Group Templates Implementation


ecolternv:
Implement the group templates feature to support creating compute domains for NVL72 support

PR #454, opened 2026-02-19
#356 - Group Template Implementation

ecolternv:
## Description
Add group templates to OSMO
- Group templates can be created and assigned to pools much like pod templates (But they cannot be assigned to individual platforms)
- Group templates will be created/deleted along with groups as part of the workflow lifecycle
- Group templates represent k8s resources that are created and destroyed along with groups submitted to a pool
- The group templates used (Kind and apiVersion) in creating a workflow are stored in the DB, so they can be correctly deleted when the workflow ends, even if the group templates for the pool change while the workflow is running

Issue #356

## Checklist

### DIFF
--- a/src/utils/job/backend_jobs.py
+++ b/src/utils/job/backend_jobs.py
@@ -463,25 +463,11 @@ class BackendSynchronizeQueues(backend_job_defs.BackendSynchronizeQueuesMixin, B
         return value
 
     def _queue_resource_api(self, api_client):
-        """Returns a dynamic resource API for the queue type.
-
-        Supports new-style specs (generic_api set) and deprecated legacy specs (custom_api or
-        resource_type only) for backwards compatibility with jobs serialized before this change.
-        """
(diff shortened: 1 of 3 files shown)


src/utils/job/backend_jobs.py
```diff
@@ -493,51 +462,47 @@ def validate_job_id(cls, value: str) -> str:
...
         return value
 
+    def _queue_resource_api(self, api_client):
+        """Returns a dynamic resource API for the queue type.
+
+        Supports new-style specs (generic_api set) and deprecated legacy specs (custom_api or
+        resource_type only) for backwards compatibility with jobs serialized before this change.
+        """
+        if self.cleanup_spec.generic_api is not None:
+            api_version = self.cleanup_spec.generic_api.api_version
```
RyaliNvidia: Maybe we can make cleanup_spec have properties for these fields to correctly fetch from generic vs custom api

PR #455, opened 2026-02-19
#407 - Authz sidecar sends accessible pools to service

RyaliNvidia:
Issue #407

## Checklist

### DIFF
--- a/deployments/charts/router/templates/_sidecar-helpers.tpl
+++ b/deployments/charts/router/templates/_sidecar-helpers.tpl
@@ -248,10 +248,6 @@ Authorization sidecar container
           name: db-secret
           key: db-password
     {{- end }}
-  ports:
-    - containerPort: {{ .Values.sidecars.authz.grpcPort }}
-      name: authz-grpc
-      protocol: TCP
   {{- if .Values.global.logs.enabled }}
   volumeMounts:
     - name: logs
--- a/deployments/charts/router/templates/log-agent-config.yaml
+++ b/deployments/charts/router/templates/log-agent-config.yaml
@@ -46,7 +46,6 @@ data:
         HTTP_Port     {{ .Values.sidecars.logAgent.prometheusPort | default 2020 }}
 
     @INCLUDE input-osmo.conf
-    @INCLUDE input-authz.conf
     @INCLUDE k8s-metadata.conf
     @INCLUDE event.conf
     @INCLUDE output-cloudwatch.conf
@@ -57,7 +56,7 @@ data:
         Tag               <service_name>
         Tag_Regex         \/var\/log\/(?<service_name>[^ ]+).txt*
         Path              /var/log/*.txt
-        Exclude_Path      /var/log/backend_event.txt* , /var/log/envoy_access_log.txt*, /var/log/envoy_api_access_log.txt*, /var/log/*authz_sidecar*.txt
+        Exclude_Path      /var/log/backend_event.txt* , /var/log/envoy_access_log.txt*, /var/log/envoy_api_access_log.txt*, /var/log/*authz_sidecar*.txt*
         Multiline         On
         Parser_Firstline  osmo-log
         Key               log
@@ -97,7 +96,6 @@ data:
         Buffer_Chunk_Size 64k
         Buffer_Max_Size   64k
 
-  input-authz.conf: |
     [INPUT]
         Name              tail
         Tag               authz_sidecar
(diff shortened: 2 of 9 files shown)


deployments/charts/router/templates/_sidecar-helpers.tpl
```diff
@@ -208,3 +208,65 @@ Log agent volumes
...
+      value: {{ .Values.services.postgres.password }}
+    {{- else if .Values.services.configFile.enabled }}
+    - name: OSMO_POSTGRES_PASSWORD
+      valueFrom:
+        secretKeyRef:
+          name: db-secret
+          key: db-password
+    {{- end }}
+  ports:
+    - containerPort: {{ .Values.sidecars.authz.grpcPort }}
```
ecolternv: We shouldn't need to add this.

Containers in a pod automatically share the same network namespace, so they are able to communicate with one another without ports. Adding the grpc port here opens it up to be connected to by things outside the pod, which is probably not what we want.

deployments/charts/service/templates/_sidecar-helpers.tpl
ecolternv, line 411: Are we listening on 0.0.0.0, or on 127.0.0.1? We should do 127.0.0.1 to prevent connections to the authz service from outside the pod

src/utils/roles/pool_access.go
```diff
@@ -0,0 +1,136 @@
...
+
+	if err := rows.Err(); err != nil {
+		return nil, fmt.Errorf("error iterating pool names: %w", err)
+	}
+
+	return names, nil
+}
+
+// GetAllowedPools evaluates role policies to determine which pools the user
+// can access based on the ActionWorkflowCreate action scoped to pool resources.
```
ecolternv, line 57: What is this x-osmo-pools header used for exactly? Is it just so we can list what pools are accessible with `osmo profile list`?

Just want to make sure I can still do `osmo pool list` and view workflows for pools that I don't have WorkflowCreate permission, but maybe have some other read-only permissions.

src/utils/roles/pool_access.go
```diff
@@ -0,0 +1,136 @@
...
+		if isPoolAllowed(userRoles, poolResource) {
+			allowed = append(allowed, poolName)
+		}
+	}
+	return allowed
+}
+
+// isPoolAllowed returns true if any role independently grants
+// ActionWorkflowCreate on the given pool resource.
+func isPoolAllowed(userRoles []*Role, poolResource string) bool {
```
ecolternv: It seems like there could be some common logic here that we could reuse for all checks. A function like

`isActionAllowedOnResource(userRoles, resource, action)`

src/utils/roles/role_cache.go
```diff
@@ -19,34 +19,108 @@ SPDX-License-Identifier: Apache-2.0
...
+func NewKeyedCache[V any](maxSize int, ttl time.Duration, logger *slog.Logger) *KeyedCache[V] {
+	return &KeyedCache[V]{
+		cache:  expirable.NewLRU[string, V](maxSize, nil, ttl),
 		logger: logger,
 	}
 }
 
+// Get retrieves a single value by key. Returns the value and true on hit.
+func (c *KeyedCache[V]) Get(key string) (V, bool) {
+	return c.cache.Get(key)
```
ecolternv, line 88: I think we should use `Peek` instead of `Get`. It seems like Get resets the TTL and peek doesnt. We could end up in a situation where we remove a role from a user, and as long as  they keep making API calls every few minutes, the cache is never cleared and so they keep that role

src/utils/roles/pool_access.go
RyaliNvidia: Yes. This only applies for profile list and some resource api which fetches information based on the user's accessible pools. osmo pool list is unchanged

deployments/charts/router/templates/log-agent-config.yaml
```diff
@@ -56,7 +57,7 @@ data:
         Tag               <service_name>
         Tag_Regex         \/var\/log\/(?<service_name>[^ ]+).txt*
         Path              /var/log/*.txt
-        Exclude_Path      /var/log/backend_event.txt* , /var/log/envoy_access_log.txt*, /var/log/envoy_api_access_log.txt*
+        Exclude_Path      /var/log/backend_event.txt* , /var/log/envoy_access_log.txt*, /var/log/envoy_api_access_log.txt*, /var/log/*authz_sidecar*.txt
```
elookpotts-nvidia: You need to ignore `/var/log/*authz_sidecar*.txt*` (note the trailing star) so that you don't pick up rotated log files as well.

elookpotts-nvidia: Same applies for the service chart.

src/utils/roles/role_cache.go
RyaliNvidia: I don't think that is the behavior. Get updates most recently used, but not ttl. TTL shouldn't be changed

deployments/charts/router/templates/log-agent-config.yaml
RyaliNvidia: Ah. Good catch

deployments/charts/service/templates/_sidecar-helpers.tpl
RyaliNvidia: Good catch. Updating

deployments/charts/router/templates/_sidecar-helpers.tpl
RyaliNvidia: I can remove this, but I'm pretty sure this doesn't expose this port to outside the pod. containerPort just serves as metadata for kubernetes to know that the port is being used and where to route if the `Service` exposes `targetPort`.

src/utils/roles/pool_access.go
RyaliNvidia: There isn't a function for this yet, but I'll make a generic function which this can use

elookpotts-nvidia (approved):

PR #457, opened 2026-02-20
Grouping Pools with Shared Nodes in the Same Nodeset

ethany-nv:
<!-- Title should be "#<issue number> - <commit title>"-->

## Description
This change groups pools with a shared node in the same nodeset, which is a construct used to determine if pools are considered shared in the UI and the CLI.

This should address the scenario depicted in the attached issue.

Cases:
```                                                                                                                                                                                                      
  - Pool A {Node1, Node2}, Pool B {Node2, Node3} → share Node2 → one nodeset
  - Pool A {Node1}, Pool B {Node2} → share nothing → two nodesets
  - Transitive: Pool A {Node1}, Pool B {Node1, Node2}, Pool C {Node2} → all three in one nodeset
  - Pool A {Node1, Node3}, Pool B {Node2, Node3} → share Node3 → one nodeset
```
Issue #458 

## Checklist

### DIFF
--- a/src/service/core/workflow/workflow_service.py
+++ b/src/service/core/workflow/workflow_service.py
@@ -250,7 +250,7 @@ def get_pool_quotas(all_pools: bool = True,
     gpu_usage_by_nodeset = {
         nodeset: sum((node_gpu_usage.get(node, NodeGpuUsage()) for node in nodeset),
                      start=NodeGpuUsage())
-        for nodeset in pools_by_nodeset.keys()
+        for nodeset in pools_by_nodeset
     }
 
     # Initialize per-pool calculated sums


src/service/core/workflow/workflow_service.py
```diff
@@ -208,15 +207,50 @@ def get_pool_quotas(all_pools: bool = True,
...
+            node_sets[pool] = node_sets[pool].add_node(
+                f'{resource.backend}/{resource.hostname}')
 
-    pools_by_nodeset = collections.defaultdict(list)
+    # Build an inverse map: node hostname -> list of pools that contain it.
+    node_to_pools: dict[str, list[str]] = collections.defaultdict(list)
     for pool, nodeset in node_sets.items():
-        pools_by_nodeset[nodeset].append(pool)
+        for node in nodeset.nodes:
+            node_to_pools[node].append(pool)
```
RyaliNvidia, line 217: Won't this create bad situations if the same name of a node exists in different clusters?

PR #467, opened 2026-02-20
Workflow API Fixes: server-side sorting, more_entries value

ethany-nv:
## Description
Fixes for Workflow API:
- Sorting order is no longer hardcoded in query
- more_entries field will return the right boolean, rather than always false

Issue - #496 

## Checklist

### DIFF
--- a/src/cli/workflow.py
+++ b/src/cli/workflow.py
@@ -991,7 +991,7 @@ def _list_workflows(service_client: client.ServiceClient, args: argparse.Namespa
             client.RequestMethod.GET,
             'api/workflow',
             params=params)
-        workflow_list += workflow_result['workflows']
+        workflow_list.extend(workflow_result['workflows'])
         current_count += count
         if args.count <= current_count or not workflow_result['more_entries']:
             break


src/service/core/workflow/helpers.py
```diff
@@ -102,7 +102,8 @@ def get_workflows(users: List[str] | None = None,
         conditions = ' AND '.join(commands)
         fetch_cmd = f'{fetch_cmd} WHERE {conditions}'
 
-    fetch_cmd += ' ORDER BY submit_time DESC LIMIT %s OFFSET %s'
+    order_direction = 'ASC' if order == connectors.ListOrder.ASC else 'DESC'
```
RyaliNvidia, line 105: This is not needed and leads to incorrect behavior. We already perform order logic here https://github.com/NVIDIA/OSMO/pull/467/changes#diff-0a18dc7ac7a629f53fa24023117e3030efba18662482b94982f164d053ec706fR110

You want to do the sorting after the limit fetching is done. The order should be applied after the first 100 (for example) entries are fetched. Not to fetch the last 100 entries

ethany-nv: This is an antipattern in API design, and this will affect our new UI as well.

A new commit has been pushed to preserve that behavior for the CLI.

src/cli/workflow.py
```diff
@@ -984,11 +991,14 @@ def _list_workflows(service_client: client.ServiceClient, args: argparse.Namespa
             client.RequestMethod.GET,
             'api/workflow',
             params=params)
-        workflow_list = workflow_result['workflows'] + workflow_list
+        workflow_list += workflow_result['workflows']
```
RyaliNvidia: Can you check if this is right? I'm pretty sure the old behavior is more correct. The current should be appended at the end. For example, If desc, it goes:
```
workflow_list: 10, 9, 8, ... 2, 1
next result is: 20, 19, 18, ... 12, 11
```
And you want 20 -> 1, not 10 -> 1 -> 20 -> 11

ethany-nv: This is correct, because in line 955 we're fixing the order to DESC. When user is specifying ASC or DESC, we always fetch DESC first.

RyaliNvidia (approved):

fernandol-nvidia (approved):

PR #473, opened 2026-02-20
database schema integration

vvnpn-nv:
## Description
Add new fields to allow schema version integration that is compatible with [pgroll](https://github.com/xataio/pgroll) for smooth database migration. This is backwards compatible as our current tables use the `public` schema.

Issue - None

## Checklist

### DIFF
--- a/src/utils/connectors/postgres.py
+++ b/src/utils/connectors/postgres.py
@@ -255,7 +255,7 @@ class PostgresConnector:
                 'Postgres Connector has not been created!')
         return PostgresConnector._instance
 
-    def _create_pool(self):
+    def _create_pool(self, search_path: str | None = None):
         """Create the ThreadedConnectionPool and semaphore."""
         try:
             if self.config.postgres_pool_minconn > self.config.postgres_pool_maxconn:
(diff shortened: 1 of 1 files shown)


src/utils/postgres/postgres_client.go
```diff
@@ -73,6 +75,14 @@ func NewPostgresClient(ctx context.Context, config PostgresConfig, logger *slog.
 	poolConfig.MinConns = config.MinConns
 	poolConfig.MaxConnLifetime = config.MaxConnLifetime
 
+	// Set search_path per connection for pgroll schema versioning
+	if config.SchemaVersion != "" && config.SchemaVersion != "public" {
+		poolConfig.AfterConnect = func(ctx context.Context, conn *pgx.Conn) error {
+			_, err := conn.Exec(ctx, "SET search_path TO "+config.SchemaVersion)
```
cypres: Looks like there is a more official way to handle this;
https://github.com/jackc/pgx/blob/49dc98edffe408f3cfa57f76bce19090ce39bd35/pgconn/config.go#L51

cypres (dismissed):

src/utils/connectors/postgres.py
cypres:
```suggestion
                options=f"-csearch_path={self.config.schema_version}"
```

src/utils/connectors/postgres.py
```diff
@@ -385,6 +394,12 @@ def __init__(self, config: PostgresConfig):
         self._init_configs()
         logging.debug('Configs initialized')
 
+        # Recreate pool with search_path set to the pgroll versioned schema
+        self._schema_initialized = True
+        if self.config.schema_version != 'public':
+            logging.debug('Switching to pgroll schema: %s', self.config.schema_version)
+            self.connect()
```
cypres, line 401: No need to reconnect?

vvnpn-nv: if it doesn't reconnect the `search_path` will still be on public

cypres (approved):

RyaliNvidia (approved):

PR #475, opened 2026-02-21
Add Pagination for Resources API

ethany-nv:
## Description
Adding pagination for resources API, to improve the user experience when querying for resources in the UI.

New parameters for `/api/resources`:

  | Parameter | Type | Default | Description |
  |-----------|------|---------|-------------|
  | `limit` | `int` | `500` | Max resources to return per page. Must be between 1 and 1000. |
  | `offset` | `int` | `0` | Number of resources to skip. Must be ≥ 0. |
  | `search` | `str` | — | Case-insensitive substring filter on node hostname (`ILIKE '%value%'`). |

  **Response additions:**

  | Field | Type | Description |
  |-------|------|-------------|
  | `has_more` | `bool` | `true` if additional resources exist beyond the current page. |

CLI is updated to work with the new changes to the limit parameter in the API. CLI will keep fetching until `has_more` becomes false.

Issue - None

## Checklist

### DIFF
--- a/src/cli/resources.py
+++ b/src/cli/resources.py
@@ -108,11 +108,23 @@ def round_resources(total_request: float, allocatable: float) -> Tuple[int, int]
 def fetch_resources(service_client: client.ServiceClient, pools: List[str],
                     platform: Optional[List[str]] = None, all_pools: bool = False) -> Dict:
     logging.debug('Getting cluster resources')
-    params = {'pools': pools, 'all_pools': all_pools}
+    base_params: Dict = {'pools': pools, 'all_pools': all_pools}
     if platform:
-        params['platforms'] = platform
+        base_params['platforms'] = platform
 
-    response = service_client.request(client.RequestMethod.GET, 'api/resources', params=params)
+    all_resources: List[Dict] = []
+    offset = 0
+    while True:
+        response = service_client.request(
+            client.RequestMethod.GET, 'api/resources',
+            params={**base_params, 'offset': offset})
+        page_resources = response.get('resources', [])
+        all_resources.extend(page_resources)
+        if not response.get('has_more', False):
+            break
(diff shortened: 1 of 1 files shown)


src/service/core/workflow/objects.py
```diff
@@ -1141,21 +1142,33 @@ def get_resources(backends: List[str] | None = None,
...
+    has_more = limit is not None and len(backend_resources) > limit
+    if has_more:
+        backend_resources = backend_resources[:limit]
+
+    return ResourcesResponse(
+        resources=[
+            workflow.ResourcesEntry.from_backend_resource(resource, verbose)
+            for resource in backend_resources
+        ],
+        has_more=has_more,
```
RyaliNvidia, line 1170: with workflow_list we call this `more_entries`. Like keep the fields the same either this or the other name

src/service/core/workflow/workflow_service.py
```diff
@@ -857,18 +857,29 @@ def get_resources(pools: List[str] | None = fastapi.Query(default = None),
...
+                  search: str | None = fastapi.Query(default=None),
                   allowed_pools_header: Optional[str] =
                     fastapi.Header(alias=login.OSMO_ALLOWED_POOLS, default=None)) -> \
     objects.ResourcesResponse | objects.PoolResourcesResponse:
     """ Returns the information of resources available in different pools. """
+    if limit < 1:
+        raise osmo_errors.OSMOUserError('Limit must be greater than 0.')
+    if limit > 1000:
+        raise osmo_errors.OSMOUserError('Limit must be at most 1000.')
+    if offset < 0:
```
RyaliNvidia, line 871: Might not need to do it in this PR, but we should create a validator that we can use everywhere rather than create this if statement in each list api

PR #499, opened 2026-02-24
pgroll database migration pre-upgrade job

vvnpn-nv:
## Description
Integrates pgroll for zero-downtime PostgreSQL schema migrations. The migration runs as a Helm pre-upgrade hook Job using postgres:15-alpine, ensuring database changes complete before new pods roll out.

Changes:
- Add migration JSON files (001-004) for the 6.0→6.2 schema migration
- Add run_migrations.sh script with contract-then-expand pattern for rollback safety
- Add Helm pre-upgrade hook Job template (migration-job.yaml) with ConfigMap auto-populated via .Files.Glob
- Add OSMO_SCHEMA_VERSION env var to service, worker, agent, router, and authz sidecar templates
- Add services.migration config section to service chart values (image, pgrollVersion, targetSchema)
- Add schemaVersion config to router chart values
- Default schema: set to public_v6_2_0 for the 6.2.0 database schema

How it works:

- Migration Job runs before pods update, applies pgroll migrations, creates versioned schema with views over physical tables
- App connects via search_path to the versioned schema (OSMO_SCHEMA_VERSION)
- Script is idempotent — safe to run multiple times, skips already-applied migrations
- To add future migrations: drop new JSON files in deployments/charts/service/migrations/

Issue - None

## Checklist

### DIFF
--- a/deployments/charts/service/values.yaml
+++ b/deployments/charts/service/values.yaml
@@ -97,7 +97,7 @@ services:
 
     ## Target pgroll schema version. Named after the app version: public_v{MAJOR}_{MINOR}_{PATCH}
     ##
-    targetSchema: public
+    targetSchema: public_v6_2_0
 
     ## Public image with bash, psql, and curl
     ##


deployments/charts/service/migrations/run_migrations.sh
```diff
@@ -0,0 +1,146 @@
+#!/usr/bin/env bash
+# Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
```
ecolternv: This copyright header needs to include apache 2.0 sdpx stuff as well.

cypres (approved):

deployments/charts/service/values.yaml
```diff
@@ -86,6 +86,58 @@ services:
...
+    ## Example for ArgoCD PreSync:
+    ##   extraAnnotations:
+    ##     argocd.argoproj.io/hook: PreSync
+    ##     argocd.argoproj.io/hook-delete-policy: BeforeHookCreation
+    ## Example for Helm pre-upgrade:
+    ##   extraAnnotations:
+    ##     "helm.sh/hook": pre-install,pre-upgrade
+    ##     "helm.sh/hook-delete-policy": hook-succeeded
+    ##
+    extraAnnotations: {}
```
ecolternv, line 127: Should we go ahead and set the hooks here or have some default? Otherwise, what is the behavior if we install this chart without any values?

ecolternv: Also, do we need a postINstall hook job to complete the migration? Otherwise there is a manual step

vvnpn-nv: updated to include the pre-upgrade hook as default

the pgroll complete actually will run the next time someone does a `helm upgrade`, in that case that it will work when someone does a patch on the existing release, also when someone goes from 6.2 -> 6.3 release

cypres (approved):

ecolternv (approved):

PR #513, opened 2026-02-24
Add token service login

xutongNV:
## Description

Issue #147

## Checklist

### DIFF
--- a/src/operator/utils/login.go
+++ b/src/operator/utils/login.go
@@ -122,8 +122,13 @@ func NewCredentials(args ListenerArgs) (credentials.PerRPCCredentials, error) {
 		return nil, fmt.Errorf("read token file %q: %w", args.TokenFile, err)
 	}
 	token := strings.TrimSpace(string(tokenBytes))
-	base := strings.TrimRight(args.ServiceURL, "/")
-	refreshURL := base + "/api/auth/jwt/access_token?access_token=" + url.QueryEscape(token)
+	baseURL, err := url.Parse(args.ServiceURL)
+	if err != nil {
+		return nil, fmt.Errorf("invalid service URL %q: %w", args.ServiceURL, err)
+	}
+	refreshURLParsed := baseURL.JoinPath("/api/auth/jwt/access_token")
+	refreshURLParsed.RawQuery = "access_token=" + url.QueryEscape(token)
+	refreshURL := refreshURLParsed.String()
 	return &TokenCredentials{
 		refreshURL: refreshURL,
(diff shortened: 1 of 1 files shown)


src/operator/utils/login.go
```diff
@@ -0,0 +1,128 @@
...
+const (
+	osmoAuthHeader = "x-osmo-auth"
+	expireWindow   = 3 * time.Second
+)
+
+// jwtExpired reports whether a JWT token's exp claim is within expireWindow of now.
+func jwtExpired(token string) bool {
+	parts := strings.Split(token, ".")
+	if len(parts) != 3 {
+		return true
```
ethany-nv, line 46: should we log that it is malformed?

src/operator/utils/login.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	payload, err := base64.RawURLEncoding.DecodeString(parts[1])
+	if err != nil {
+		return true
```
ethany-nv, line 51: could log the error?

src/operator/utils/login.go
```diff
@@ -0,0 +1,128 @@
...
+	"strings"
+	"sync"
+	"time"
+
+	"google.golang.org/grpc/credentials"
+)
+
+const (
+	osmoAuthHeader = "x-osmo-auth"
+	expireWindow   = 3 * time.Second
```
ethany-nv: it actually expires in 5, right? maybe we can set this to 4?

src/operator/utils/login.go
```diff
@@ -0,0 +1,131 @@
...
+func NewCredentials(args ListenerArgs) (credentials.PerRPCCredentials, error) {
+	if args.TokenFile == "" {
+		return nil, nil
+	}
+	tokenBytes, err := os.ReadFile(args.TokenFile)
+	if err != nil {
+		return nil, fmt.Errorf("read token file %q: %w", args.TokenFile, err)
+	}
+	token := strings.TrimSpace(string(tokenBytes))
+	base := strings.TrimRight(args.ServiceURL, "/")
```
RyaliNvidia: You can maybe use paths.Join instead of trimming for combining url paths: https://stackoverflow.com/questions/34668012/combine-url-paths-with-path-join

RyaliNvidia: This seems like a general utils function that could be helpful across the repo

xutongNV: keep trimming token, as token is reading from a file, and sometimes a new empty end line may be added to the file

PR #515, opened 2026-02-24
Fix user role map syncing and pool mapping in authz service

RyaliNvidia:
Issue #None

## Checklist

### DIFF
--- a/src/utils/roles/action_registry.go
+++ b/src/utils/roles/action_registry.go
@@ -1132,10 +1132,23 @@ func CheckSemanticAction(
 	pgClient *postgres.PostgresClient,
 ) AccessResult {
 	resolvedAction, resolvedResource := ResolvePathToAction(ctx, path, method, pgClient)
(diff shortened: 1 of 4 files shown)


cypres (approved):

src/utils/roles/action_registry.go
```diff
@@ -1023,6 +1023,7 @@ func ConvertRoleToSemantic(role *Role) *Role {
 
 	for _, policy := range role.Policies {
 		var semanticActions []RoleAction
+		seenActions := make(map[string]bool)
```
cypres: FYI, the idiomatic way to make a set in Go is like this
```
s := make(map[string]struct{})
s["foo"] = struct{}{}
_, hasFoo := s["foo"]
```

Because an empty struct in Go does not involve memory allocation

RyaliNvidia: Oh interesting. I'll update it to this. Thanks

cypres (dismissed):

vvnpn-nv (dismissed):

Issue #377, opened 2026-02-05
Add Golang Metrics Library through OpenTelemetry


xutongNV:
Our service uses OpenTelemetry to write metrics to Prometheus. Previously, we have a Python library that our Python services leverage to write metrics, but as we start migrating to Golang, we need a new Golang version of that library to help write the metrics from our new Golang services.

PR #517, opened 2026-02-25
Backend Messages Queue Metric from Message Worker

ethany-nv:
## Description
Allow message worker to track the backend messages queue length periodically.

Issue #377 

## Checklist

### DIFF
--- a/src/service/agent/message_worker.py
+++ b/src/service/agent/message_worker.py
@@ -90,13 +90,14 @@ class MessageWorker:
         # Progress writer for liveness/readiness probes
         self._progress_writer = progress.ProgressWriter(config.progress_file)
 
-        # Register observable gauge for stream queue length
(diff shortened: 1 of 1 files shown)


src/service/agent/message_worker.py
```diff
@@ -88,11 +90,26 @@ def __init__(self, config: MessageWorkerConfig):
         # Progress writer for liveness/readiness probes
         self._progress_writer = progress.ProgressWriter(config.progress_file)
 
+        if config.method != 'dev':
```
xutongNV, line 93: this metric is pushed one time on init right? Is this useful?

ethany-nv: No, this metric is pushed periodically (once a minute). Take a look at worker.py, we do something very similar there.

PR #549, opened 2026-02-26
draft: 6.2 upgrade guide

vvnpn-nv:
## Description

Issue #<issue number | "None">

## Checklist
- [ ] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/OSMO/blob/main/CONTRIBUTING.md).
- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.

### DIFF
--- a/deployments/6_2_upgrade_guide.md
+++ b/deployments/6_2_upgrade_guide.md
@@ -1,0 +1,357 @@
+# Upgrading OSMO from 6.0 to 6.2
+
+## What's new in 6.2
+
+- **New authentication architecture** — oauth2Proxy sidecar + authz sidecar replace the old Envoy-native oauth2Filter
+- **RBAC system** — new database tables for users, roles, and role mappings managed by the authz sidecar
+- **pgroll database migrations** — zero-downtime schema changes via versioned schemas
+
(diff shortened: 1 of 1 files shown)


deployments/6_2_upgrade_guide.md
```diff
@@ -0,0 +1,357 @@
+# Upgrading OSMO from 6.0 to 6.2
```
RyaliNvidia, line 1: NIT: copyright

deployments/6_2_upgrade_guide.md
```diff
@@ -0,0 +1,357 @@
...
+The file must contain (plain text, not base64):
+
+```
+client_secret = "<value>"
+cookie_secret = "<value>"
+```
+
+`cookie_secret` must be exactly 16, 24, or 32 raw bytes. When using Vault agent, include this template in **both** `config.hcl` and `config-init.hcl` to avoid a race condition.
+
+### Step 4: Add authz sidecar (service and router charts only)
```
RyaliNvidia, line 233: You don't really need to specify this since this is the default behavior

PR #588, opened 2026-03-02
fix: allow oauth2 proxy to redirect browser to a logout endpoint

elookpotts-nvidia:
## Description

In the UI, signing out did not fully log users out — after clicking "Sign out," users were immediately redirected back to the app with an active session because OAuth2 Proxy cleared its own cookie but the identity provider's browser-side SSO session remained alive, triggering silent re-authentication. The fix adds a configurable `oidcEndSessionUrl` value and an Envoy Lua filter that injects a `rd=<oidcEndSessionUrl>` redirect parameter on `/oauth2/sign_out` requests, causing the browser to visit the IdP's end-session endpoint after the local session is cleared. A companion `--whitelist-domain` entry in `extraArgs` is required to allow OAuth2 Proxy to honor the external redirect target.

Issue #None

## Checklist

### DIFF
--- a/deployments/charts/web-ui/templates/_envoy-config-helpers.tpl
+++ b/deployments/charts/web-ui/templates/_envoy-config-helpers.tpl
@@ -155,7 +155,7 @@ virtual_hosts:
   routes:
   {{- if $.Values.sidecars.oauth2Proxy.enabled }}
   - match:
-      path: /oauth2/idp_sign_out
+      path: /signout
     redirect:
       {{- if $.Values.sidecars.oauth2Proxy.oidcEndSessionUrl }}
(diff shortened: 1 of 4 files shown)


fernandol-nvidia (dismissed):

deployments/charts/web-ui/templates/_envoy-config-helpers.tpl
```diff
@@ -154,6 +154,14 @@ virtual_hosts:
   domains: ["*"]
   routes:
   {{- if $.Values.sidecars.oauth2Proxy.enabled }}
+  - match:
+      path: /oauth2/idp_sign_out
```
cypres: Should we just make this `/signout` since it's more memorable?

elookpotts-nvidia: Works for me!

PR #635, opened 2026-03-06
add heartbeat message

xutongNV:
## Description

Unlike websocket, ping/pong messages for stream rpc isn't considered as activities by ALB, connections will be idle timed out by ALB after 60 seconds.

Application level of heartbeat messages from both direction keep the connection alive, so backend can receive node condition config change notifications from service.

This PR also make a envoy config change that fixes an issue:
Envoy's ext_authz filter is configured with with_request_body: max_request_bytes: 8192, which
   buffers request body data before sending the authorization check. For gRPC streaming RPCs, the HTTP/2
  END_STREAM flag is never set, so envoy waits until 8192 bytes accumulate. Streams that don't send enough 
  data quickly (like NodeConditionStream and idle ListenerStreams) never reach that threshold, so the auth
  check never fires and the request hangs forever — never reaching the backend.

Issue #147

## Checklist

### DIFF
--- a/deployments/charts/service/templates/_envoy-config.tpl
+++ b/deployments/charts/service/templates/_envoy-config.tpl
@@ -351,11 +351,6 @@ data:
                 typed_config:
                   "@type": type.googleapis.com/envoy.extensions.filters.http.ext_authz.v3.ExtAuthz
                   transport_api_version: V3
-                  {{- if not $useHttp2 }}
-                  with_request_body:
-                    max_request_bytes: 8192
-                    allow_partial_message: true
-                  {{- end }}
                   failure_mode_allow: false
                   grpc_service:
                     envoy_grpc:


cypres (approved):

deployments/charts/service/templates/_envoy-config.tpl
```diff
@@ -351,9 +351,11 @@ data:
                 typed_config:
                   "@type": type.googleapis.com/envoy.extensions.filters.http.ext_authz.v3.ExtAuthz
                   transport_api_version: V3
+                  {{- if not $useHttp2 }}
```
cypres: We might be able to just remove the request body from authz, @RyaliNvidia ?

xutongNV: Yeah we briefly talked about it as well. stream case doesnt need the data to authz

RyaliNvidia: Yeah. re-evaluating the sidecar, we currently do not need the body. We can add it back in if we add an api which requires it

deployments/charts/service/templates/_envoy-config.tpl
```diff
@@ -351,9 +351,11 @@ data:
...
                   with_request_body:
                     max_request_bytes: 8192
                     allow_partial_message: true
+                  {{- end }}
```
cypres: You can just remove the with_request_body entirely, we don't currently read it.
```suggestion
```

xutongNV: @RyaliNvidia can you confirm this please?

RyaliNvidia: Yes we can remove this. I thought we were using this, but we weren't and just had plans for future apis. We can reintroduce this later when an api needs it

cypres (approved):

PR #642, opened 2026-03-09
Resolve issues creating default admin

RyaliNvidia:
Our validation logic broke the 10 year allowance on the admin password + we didn't verify that the password was of the correct length

Issue #None

## Checklist

### DIFF
--- a/docs/deployment_guide/appendix/authentication/index.rst
+++ b/docs/deployment_guide/appendix/authentication/index.rst
@@ -125,6 +125,10 @@ The service receives these via:
 - **Args:** ``--default_admin_username <username>`` when ``defaultAdmin.enabled`` is true.
 - **Env:** ``OSMO_DEFAULT_ADMIN_PASSWORD`` from the secret ``passwordSecretName`` and ``passwordSecretKey``.
 
+.. important::
+
+   The admin password must be 43 characters long.
+
 Example: create the secret and enable default admin in your values:
 
 .. code-block:: bash


docs/deployment_guide/appendix/authentication/index.rst
```diff
@@ -125,6 +125,10 @@ The service receives these via:
 - **Args:** ``--default_admin_username <username>`` when ``defaultAdmin.enabled`` is true.
 - **Env:** ``OSMO_DEFAULT_ADMIN_PASSWORD`` from the secret ``passwordSecretName`` and ``passwordSecretKey``.
 
+.. important::
+
+   The admin password must be 43 characters long.
```
vvnpn-nv, line 130: The password has to be exactly 43 chars?

xutongNV (approved):

docs/deployment_guide/appendix/authentication/index.rst
adelbertc: I'm guessing it's b/c it passes through code that treats it as a refresh token. But that begs the question why we are checking for refresh token string **length** rather than something more semantic than that?

RyaliNvidia: Yeah. The use of this "admin" user is only in setting up the environment where there isn't an IdP and not really for a production environment. This then gets treated as an access token which we enforce the length.

We don't need to verify anything more than the string length in the jwt fetch because normal access token generation is from the "random" `secrets.token_urlsafe` which ensures that there is alreay enough entropy from the token generation. We only need to enforce the length so it reduces the collision space from brute force attacks.

Otherwise, validating that the properties of this password isn't really required since we don't really have the strict use case right now where we are handling multiple users logging into the OSMO in a production environment without an IdP.

adelbertc: Gotcha.

adelbertc (approved):

PR #665, opened 2026-03-10
Skip stale resource messages in multi-worker processing

xutongNV:
## Description

- Add `last_updated` and `last_usage_updated` TIMESTAMP columns to the `resources` table
- Use SQL `WHERE` clauses in `update_resource`, `update_resource_usage`, and `clean_resources` to atomically skip stale writes when multiple message workers process messages out of order from the Redis Stream
- Pass `message_timestamp` from both the message worker (Redis Stream path) and websocket path (`backend_listener_impl`)
- Deletes (node removal) are unguarded — they always proceed regardless of timestamp

Issue #147

## Checklist

### DIFF
--- a/src/service/agent/helpers.py
+++ b/src/service/agent/helpers.py
@@ -267,17 +267,19 @@ def update_resource(backend: str, message: backend_messages.UpdateNodeBody,
     rowcount = postgres.execute_commit_command(commit_cmd, columns)
     if rowcount > 0:
         pool_config = connectors.fetch_verbose_pool_config(postgres, backend)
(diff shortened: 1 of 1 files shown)


deployments/charts/service/migrations/005_v6_3_0_schema.json
```diff
@@ -0,0 +1,24 @@
+{
```
RyaliNvidia, line 1: I think we would need to name the file like `005_v6_3_0_schema.json`, but this is fine is we want to leave the version undetermined

vvnpn-nv (dismissed):

PR #649, opened 2026-03-09
Assign unique Prometheus metrics ports per service in bazel mode

cypres:
## Description
Prevents port conflicts when running all services locally via bazel. Core: 9464, worker: 9465, delayed_job_monitor: 9466, backend_listener: 9467, backend_worker: 9468.

Issue #633

Tested with `bazel run @osmo_workspace//run:start_service -- --mode bazel`

## Checklist
- [X] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/OSMO/blob/main/CONTRIBUTING.md).
- [X] New or existing tests cover these changes.
- [X] The documentation is up to date with these changes.

PR #673, opened 2026-03-10
prometheus: redo fix by using dev mode

cypres:
## Description
Redo of #670 and #649 

Upon closer examination we would actually prefer not running the prometheus server in dev mode, which means we can avoid overriding the configuration in each service.

This reverts the previous commits and fixes it by just not running prometheus servers in dev mode.

Issue #649

## Checklist

## Summary by CodeRabbit

* **Infrastructure**
  * Standardized metrics collection port to 9464 across backend listener, backend worker, delayed job monitor, and worker services
  * Removed custom Prometheus metrics port configuration
  * Disabled OpenTelemetry metrics collection by default


RyaliNvidia (dismissed):

xutongNV (commented):
can we set defult to false

fernandol-nvidia (dismissed):

xutongNV (approved):

PR #678, opened 2026-03-11
Rudimentary secret redaction

cypres:
## Description
Rudimentary secret redaction in case a user puts a secret into the workflow spec directly.

Issue #None

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Workflow specs are now streamed with automatic secret redaction — sensitive credentials are masked while preserving non-sensitive content and original lengths.

* **Tests**
  * Added tests validating redaction for plaintext and base64-encoded secrets, ensuring safe content is preserved and masking behaves consistently.

### DIFF
--- a/src/service/core/workflow/workflow_service.py
+++ b/src/service/core/workflow/workflow_service.py
@@ -65,18 +65,10 @@ _BASE64_FRAGMENT_RE = re.compile(r'(?<![A-Za-z0-9+/])[A-Za-z0-9+/]{16,}={0,2}(?!
 
 def redact_secrets(lines: Iterable[str]) -> Generator[str, None, None]:
     """ Yield lines with secrets in the spec redacted. """
-    def redact_value(secret: str) -> str:
-        """ Replace a secret with **redacted**, padded with '*' to match original length. """
-        replacement = '**redacted**'
-        padding = max(0, len(secret) - len(replacement))
-        left = padding // 2
-        right = padding - left
-        return '*' * left + replacement + '*' * right
-
     def redact_base64_fragments(line: str) -> str:
         """
         Find base64-encoded fragments in a line, decode them, redact any secrets found inside,
-        and re-encode the redacted content back to base64.
+        and replace the whole fragment with [MASKED].
         """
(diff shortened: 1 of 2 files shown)


src/service/core/workflow/workflow_service.py
```diff
@@ -48,6 +49,56 @@
...
+
+def redact_secrets(lines: Iterable[str]) -> Generator[str, None, None]:
+    """ Yield lines with secrets in the spec redacted. """
+    def redact_value(secret: str) -> str:
+        """ Replace a secret with **redacted**, padded with '*' to match original length. """
+        replacement = '**redacted**'
+        padding = max(0, len(secret) - len(replacement))
+        left = padding // 2
+        right = padding - left
+        return '*' * left + replacement + '*' * right
```
RyaliNvidia: NIT: I think this can be a default specified length instead of matching the secret length. I know in logs, we do `[[MASKED]]`

ethany-nv: I agree, fixed length is cleaner and simpler

src/service/core/workflow/workflow_service.py
```diff
@@ -48,6 +49,56 @@
 
 FETCH_TASK_LIMIT = 1000
 
+# Regex to match secrets in the spec. While this is not a perfect solution, it solves the majority
```
ethany-nv, line 52: So your MR addresses leaking of secrets in a script that we inline in our workflow, but users have also directly put their keys in the environment variables section of the spec:

```
workflow:
  name: train-with-checkpointing
  tasks:
  - name: train-with-checkpointing
    image: ubuntu:24.04
    command: [/bin/bash]
    environment:
      WANDB_KEY: <my key>
```

Maybe you wanna address this too in your PR? Or you can address it in a followup

cypres: It's actually already redacted.

xutongNV: this one is just a key name to fetch the real credential, it doesnt really matter

PR #681, opened 2026-03-11
fix backend listener

xutongNV:
## Description

Issue #None

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved time handling by switching internal timestamp parsing and timeout calculations to timezone-aware UTC. This makes container error timeout detection and related status reporting consistent and reliable across different system timezones, reducing false or inconsistent backend error markings.

### DIFF
--- a/src/operator/tests/test_pod_status_calculator.py
+++ b/src/operator/tests/test_pod_status_calculator.py
@@ -34,18 +34,19 @@ from src.utils.job import task
 
 
 def parse_time_string(time_str: str) -> datetime.datetime:
-    """Parse time strings like 'now', 'now-5m', 'now-15m' into datetime objects"""
+    """Parse time strings like 'now', 'now-5m', 'now-15m' into timezone-aware UTC datetimes."""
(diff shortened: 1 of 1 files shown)


src/operator/backend_listener.py
```diff
@@ -930,7 +930,7 @@ def calculate_pod_status(pod: Any) -> Tuple[task.TaskGroupStatus, str, Optional[
                     # When a container fails to create, the pod will not be Ready.
                     # The lastTransitionTime of this condition is the closest timestamp.
                     if condition.type == 'Ready' and condition.status == 'False':
-                        now = datetime.datetime.now()
+                        now = datetime.datetime.now(datetime.timezone.utc)
```
ethany-nv, line 933: anywhere else in listener where we call datetime.datetime.now()?

ethany-nv (commented):
Please fix title and description

src/operator/backend_listener.py
xutongNV: no

PR #692, opened 2026-03-11
Migrate NodeListener from bidi streaming to unary RPC

xutongNV:
Add SendListenerMessage unary RPC to ListenerService, enabling NodeListener to use simple request-reply instead of bidirectional streaming. This improves AWS ALB compatibility (no HTTP/2 PING needed), simplifies the code (2 goroutines instead of 3, no unacked message queue), and enables better load balancing.

Changes:
- Proto: add SendListenerMessage RPC to ListenerService
- Server: add SendListenerMessage handler with Redis push and ACK
- BaseListener: add RunUnary (2-goroutine lifecycle), SendUnaryMessage, and gRPC built-in retry config for transient failures
- NodeListener: switch from Run/SendMessage to RunUnary/SendUnaryMessage
- Tests: add coverage for unary handler, base listener RunUnary, and node listener sendMessages paths

## Description

Issue #147 

## Checklist

### DIFF
--- a/src/operator/event_listener.go
+++ b/src/operator/event_listener.go
@@ -33,7 +33,7 @@ import (
 	pb "go.corp.nvidia.com/osmo/proto/operator"
 )
 
-// EventListener manages the bidirectional gRPC stream connection for Kubernetes events
+// EventListener manages unary RPC calls for Kubernetes events
 type EventListener struct {
 	*utils.BaseListener
 	args utils.ListenerArgs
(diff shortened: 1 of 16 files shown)


cypres (approved):
This certainly simplifies things from a networking point of view.

At large scale of Unary RPC/s there is an advantage to batching up the ListenerMessages into a repeated type to reduce overhead, but that of course requires us to handle that buffering/batching client side, so that is perhaps best for a later optimization.

Issue #661, opened 2026-03-10
We should surface HEAD errors (e.g. 4XX, 5XX) before streaming file responses in APIs


fernandol-nvidia:
## Issue

When file streaming (e.g. `/spec`, `/events`, `/logs`) fail due to legitimate issues (e.g. 4XX, 5XX), we currently do not catch early and return to users.

Instead, we attempt to stream the file anyways only for the stream to break (due to the aforementioned issues).

## Proposal

Rather than attempting to stream immediately, we should catch any `OSMODataStorageClientError` at: https://github.com/NVIDIA/OSMO/blob/main/src/service/core/workflow/helpers.py#L354-L363

Before we stream anything, we always make a HEAD call first to ensure the file is accessible, for example:

https://github.com/NVIDIA/OSMO/blob/main/src/lib/data/storage/backends/s3.py#L727
https://github.com/NVIDIA/OSMO/blob/main/src/lib/data/storage/backends/azure.py#L442

This gives us the opportunity to "pre-validate" and surface any *real* and meaning errors back to the client.

### Minimum reproducible example

```shell
1. Have invalid S3 credentials (no read scope, expired, etc)
2. Attempt to fetch logs/events/spec
```

### Relevant log output

```shell
Error message: Failed to fetch logs: Response ended prematurely
Error code: 1
```

### Code of Conduct

- [x] I agree to follow NVIDIA OSMO's Code of Conduct
- [x] I have searched the [open bugs](https://github.com/NVIDIA/OSMO/issues?q=is%3Aopen+is%3Aissue+label%3Abug) and have found no duplicates for this bug report

PR #730, opened 2026-03-20
Handle 403 in streaming response when fetching workflow file

ethany-nv:
## Description
When cloud storage credentials are invalid, streaming endpoints (logs, events, spec, error_logs) return HTTP 200 then abruptly terminate. Clients see ChunkedEncodingError: Response ended prematurely instead of a meaningful error message.

StreamingResponse commits the 200 status before iterating the lazy storage generator. When the actual storage call fails (e.g. 403), the exception can't be converted to an HTTP error — the stream just dies.

Pre-validate storage access before streaming:
  - helpers.get_workflow_file() now calls workflow_file_exists() before returning the stream, eagerly hitting the storage backend. Credential/permission errors raise before StreamingResponse is constructed, returning a proper HTTP 500 with a JSON error body.
  - Registered OSMODataStorageError in the FastAPI exception handler chain.

Show the actual error message instead of a generic status code:
  - Added parseStreamErrorResponse() helper that reads the JSON error body from failed streaming responses.
  - Updated useLogStream and useEventStream hooks to extract and display the backend error message (e.g. "Internal server error: 403") instead of "Stream failed: 500 Internal Server Error".

Issue #661 

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Added top-level handling for a specific storage error to prevent unhandled exceptions.
  * Workflow file access now performs an explicit pre-check to surface backend/permission issues earlier.
  * Streaming hooks now propagate parsed error content instead of generic status text.

* **New Features**
  * Added async stream-error parsing to produce clearer, user-facing error messages from streamed responses.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -85,3 +85,7 @@ build:ci --nokeep_state_after_build
 # Testcontainers Environment
 build:ci --test_env=DOCKER_HOST=tcp://docker:2375
 build:ci --test_env=TESTCONTAINERS_HOST_OVERRIDE=docker
+
+# Coverage: use with `bazel coverage`
+coverage --combined_report=lcov
+coverage --test_tag_filters=-nocoverage
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -1,0 +1,195 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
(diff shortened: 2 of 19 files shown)


cypres (approved):

src/service/core/workflow/helpers.py
```diff
@@ -310,6 +310,28 @@ def get_pool_resources(pools: List[str] | None = None,
...
+
+    Returns the stream wrapped in error handling so that storage errors
+    (e.g. invalid credentials returning 403) are yielded as text content
+    instead of raising an exception that kills the HTTP streaming response.
+    """
+    def _inner() -> Generator[str, None, storage.StreamSummary]:
+        try:
+            yield from stream
+        except Exception as error:  # pylint: disable=broad-exception-caught
+            yield f'\nError: Failed to retrieve {file_name}: {error}\n'
```
fernandol-nvidia: This will result in a 200 OK response with the content being `Error: Failed to retrieve ...`.

Which is very different from a 403 Unauthorized response.

src/service/core/workflow/helpers.py
```diff
@@ -352,14 +374,20 @@ def get_workflow_file(file_name: str, workflow_name: str,
             )
 
     if last_n_lines is not None:
-        return storage_client.get_object_stream(
-            file_prefix,
-            last_n_lines=last_n_lines,
+        return _stream_with_error_handling(
+            storage_client.get_object_stream(
```
fernandol-nvidia: storage_client.get_object_stream will make a `HEAD` first before streaming anything. So we should try/catch `storage_client.get_object_stream` and return a https://fastapi.tiangolo.com/tutorial/handling-errors/#import-httpexception.

Issue #646, opened 2026-03-09
Investigate and Address Inefficient N+1 Queries to Postgres in Workers


fernandol-nvidia:
### What new feature or improvement would you like to see?

In various places of our codebase, we have N+1 patterns when querying DB.

N+1 query pattern means that we have N tasks, and we make 1 DB query per task, resulting in N roundtrips, N transactions, N DB read/writes. 

For better throughput and atomicity, we should strive for transactions.

This applies to both Python and Golang

### Why is this needed?

Improve throughput and consistency.

Avoiding DOS during heavy load.

### Code of Conduct

- [x] I agree to follow NVIDIA OSMO's Code of Conduct
- [x] I have searched the [open feature requests](https://github.com/NVIDIA/OSMO/issues?q=is%3Aopen+is%3Aissue+label%3Afeature) and have found no duplicates for this feature request

Issue #411, opened 2026-02-10
Supporting Submission of Large Workflows


ethany-nv:
### What new feature or improvement would you like to see?

Support workflows with a CreateGroup job spec that is larger than 10MB.

### Why is this needed?

Currently, there is a benchmarking workflow that works on exactly 100 tasks, but the websocket connection between service and backend worker starts breaking once it the workflow has 101 tasks or more. We are working with customers that require their workflows to support up to 400 tasks (potentially a job spec of 40MB), and this spec needs to be passed seamlessly from service to backend worker without any issues.

### Code of Conduct

- [x] I agree to follow NVIDIA OSMO's Code of Conduct
- [x] I have searched the [open feature requests](https://github.com/NVIDIA/OSMO/issues?q=is%3Aopen+is%3Aissue+label%3Afeature) and have found no duplicates for this feature request

PR #742, opened 2026-03-24
Optimize UpdateGroup DB performance: batch updates, lightweight aggregation, skip unnecessary task loading

ethany-nv:
## Description

`UpdateGroup` job still makes inefficient DB calls and Pydantic object constructions.

This PR improves the following:

  - Batch sibling task updates: Replace N individual UPDATE tasks calls with a single batch UPDATE when setting sibling tasks to terminal status on group cancel/finish. Reduces N DB round-trips to 1 per group event.
  - Lightweight status aggregation: Replace SELECT t.*, workflows.workflow_uuid ... JOIN workflows (N full rows + N Pydantic object constructions) with SELECT status, lead, COUNT(*) GROUP BY status, lead (~2-4 rows, no JOIN) when computing group status in TaskGroup.update_status_to_db().
  - Skip unnecessary task loading: Add TaskGroup.fetch_metadata_from_db() which skips the Task.list_by_group_name() query. 5 call sites that only need group-level metadata (status, name, spec, group_uuid) now use it instead of loading all task rows.

Index to Create:
```
  CREATE INDEX IF NOT EXISTS tasks_workflow_group_name_retry_idx
      ON tasks (workflow_id, group_name, name, retry_id);
```

Issue #411 #646 

## Checklist

## Summary by CodeRabbit

* **Refactor**
  * Bulk task status updates and DB-side status aggregation for more reliable, faster group state transitions.
  * Group retrieval now uses lightweight metadata loading to avoid unnecessary task data fetches.

* **Tests**
  * Added comprehensive unit tests for aggregation and batch-update validation; strengthened test helpers and typing.
  * Added Postgres-backed integration tests and a new test target to validate DB behavior for group/task status updates.

### DIFF
--- a/src/utils/job/tests/test_task_db.py
+++ b/src/utils/job/tests/test_task_db.py
@@ -1,5 +1,5 @@
 """
-SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -106,11 +106,14 @@ class TaskDbFixture(
              json.dumps({}), lead))
         return task_db_key
 
-    def _fetch_task_status(self, task_name: str, retry_id: int = 0) -> dict:
+    def _fetch_task_status(self, task_name: str, retry_id: int = 0,
+                           group_name: str = GROUP_NAME) -> dict:
+        fetch_cmd = '''
+            SELECT status, end_time, failure_message, exit_code FROM tasks
+            WHERE workflow_id = %s AND group_name = %s AND name = %s AND retry_id = %s
+        '''
         rows = self._get_db().execute_fetch_command(
-            '''SELECT status, end_time, failure_message, exit_code FROM tasks
-               WHERE workflow_id = %s AND name = %s AND retry_id = %s''',
-            (WORKFLOW_ID, task_name, retry_id), True)
+            fetch_cmd, (WORKFLOW_ID, group_name, task_name, retry_id), True)
         return rows[0]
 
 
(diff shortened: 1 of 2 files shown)


src/utils/job/tests/test_task_db.py
```diff
@@ -0,0 +1,301 @@
+"""
+SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
RyaliNvidia:
```suggestion
SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

RyaliNvidia: Please check the rest of the files

src/utils/job/jobs.py
```diff
@@ -704,15 +704,20 @@ def _update_all_tasks(
...
-                    # (not FAILED_EVICTED)
-                    status = task.TaskGroupStatus.FAILED if self.status.failed() else self.status
-                    # TODO: Add a new status type for status caused by Lead Container finishing
-                    task_obj.update_status_to_db(update_time, status, 'Lead task finished')
+                # If group leader exits with a special failed status like
+                # FAILED_EVICTED, the other tasks should be labeled as FAILED
+                # (not FAILED_EVICTED)
+                sibling_status = task.TaskGroupStatus.FAILED if self.status.failed() \
+                    else self.status
+                task.Task.batch_update_status_to_db(
```
xutongNV, line 712: will tasks' retry id increase during the update duration?

ethany-nv: At line 712, this code path runs when the lead task finished with a non-RESCHEDULED status (line 706 else branch). The lead task is done, so no new retries for the lead would be created. For sibling tasks, they're being set to terminal status because the lead finished — no retry would be happening for them either at this point. The lead_task_name exclusion skips the lead task, and the siblings haven't been rescheduled (that only happens in the RESCHEDULED path at lines 688-705).

Issue #800, opened 2026-04-03
Run a Workflow Locally


maufrancom:
### What new feature or improvement would you like to see?

A new osmo local run CLI subcommand that executes OSMO workflow specification YAMLs locally using Docker containers, bypassing the Kubernetes control plane entirely. This eliminates the need for a running OSMO service, a Kubernetes cluster, Redis, PostgreSQL, or any other platform infrastructure.

### Why is this needed?

We want internal developers, partners and the external community to be able to reproduce an OSMO workflow in their local machines as efficiently as possible.  This would standardize all physical AI workflows (Cosmos, NuRec, Isaac, etc) on OSMO regardless of the capabilities of the user to run them.

### Code of Conduct

- [x] I agree to follow NVIDIA OSMO's Code of Conduct
- [x] I have searched the [open feature requests](https://github.com/NVIDIA/OSMO/issues?q=is%3Aopen+is%3Aissue+label%3Afeature) and have found no duplicates for this feature request

PR #802, opened 2026-04-03
#800 - Local workflow execution

maufrancom:
## Description
A new osmo local run CLI subcommand that executes OSMO workflow specification YAMLs locally using Docker containers, bypassing the Kubernetes control plane entirely. This eliminates the need for a running OSMO service, a Kubernetes cluster, Redis, PostgreSQL, or any other platform infrastructure.

Issue #800

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added standalone and Docker‑Compose local runners plus new CLI commands to run/resume workflows with variable overrides, credential mappings, shm-size, GPU support, and optional work‑dir retention.

* **Tests**
  * Extensive unit and Docker‑gated integration tests for both runners covering scheduling, token substitution, GPU behavior, filesystem safety, compose YAML generation, and error conditions.

* **Documentation**
  * Updated docs describing standalone and compose modes, networking, GPU and credential behaviors.

* **Chores**
  * Added tutorial spec aggregation, build/test targets, and updated .gitignore.


vvnpn-nv (commented):
Thanks for the contribution, the motivation is solid and we should support it. However, there are some architectural concerns worth discussing before merging.

Key Issues
1. Container paths diverge from on-cluster behavior. On-cluster, `{{output}}` resolves to `/osmo/data/output` and `{{input:0}}` to `/osmo/data/input/0`. This PR mounts at host-absolute paths (`/tmp/osmo-local-abc123/task1/output`), so scripts that work locally may break on-cluster (or vice versa).

2. "Parallel" tasks run sequentially. `_find_ready_tasks()` returns all independent tasks, but the `for` loop executes them one at a time and bails on first failure. This masks race conditions and makes execution order non-deterministic.

3. No cycle detection. Circular dependencies produce zero ready tasks. The loop never enters,` _results` is empty, and `execute()` returns True — reporting success when nothing ran.

4. Task group semantics are lost. Groups are flattened to individual tasks, dropping barrier synchronization, lead-task failure policy (`ignoreNonleadStatus`), and inter-task DNS (`{{host:taskname}}`).

5. `--docker` flag is silently ignored. Accepted in argparse but never passed to `run_workflow_locally()` or `LocalExecutor`.

6. Unresolved Jinja variables pass through silently. Template detection checks for `{%%`, `{#`, default-values but misses bare `{{ variable }}` without a `default-values:` section.

**Suggestion on an alternative solution to solve this problem:**
Rather than a custom executor reimplementing scheduling, consider generating a `docker-compose.yml` and running `docker compose up` behind the scenes. This gives you:

* Correct paths:  bind-mount at `/osmo/data/output` and `/osmo/data/input/N:ro`, matching on-cluster behavior
* Real parallel execution and native dependency ordering via `depends_on: { condition: service_completed_successfully }`
* Cycle detection for free (Compose validates the DAG)
* Group DNS — Docker Compose service names are DNS-addressable, so `{{host:taskname}}` just works
* No custom scheduler to maintain — Compose handles lifecycle, logs, cleanup, and failure propagation

maufrancom:
@coderabbitai resume

maufrancom:
@coderabbitai resume

Issue #893, opened 2026-04-23
deploy-k8s.sh diverges from project conventions: 3 independent failures block a clean deploy-osmo-minimal.sh run


yoosful:
### Describe the bug.

Running the project's own one-shot installer on a fresh Kubernetes cluster hits three independent failures, all traceable to the same underlying pattern: `deployments/scripts/deploy-k8s.sh` installs the sub-charts (`osmo/service`, `osmo/backend-operator`) individually and diverges from the conventions established by both the `quick-start` umbrella chart and the documented procedure in [`docs/deployment_guide/appendix/deploy_minimal.rst`](https://github.com/NVIDIA/OSMO/blob/6.2.10/docs/deployment_guide/appendix/deploy_minimal.rst).

**Affected versions.** `6.2.10` (latest released) and `main @ 22762c7`. `deploy-k8s.sh` is byte-identical between these two revisions (only one unrelated commit in between).

Filing three sub-bugs in the same issue since they share one file and one root cause — they can all be addressed in a single PR.

---

### Bug 1 — Helm install fails on clusters without the Prometheus Operator

The auto-generated values files written by `deploy-k8s.sh` set `sidecars.otel.enabled: false` but do not set `podMonitor.enabled: false`. The sub-chart defaults are `enabled: true`:

- [`deployments/charts/service/values.yaml:1192`](https://github.com/NVIDIA/OSMO/blob/6.2.10/deployments/charts/service/values.yaml#L1192) — renders `templates/api-monitor.yaml`
- [`deployments/charts/backend-operator/values.yaml:230`](https://github.com/NVIDIA/OSMO/blob/6.2.10/deployments/charts/backend-operator/values.yaml#L230) — renders `templates/api-monitor.yaml`

The `backend-operator` chart even has a comment at `values.yaml:227` stating _"Set enabled: false if the PodMonitor CRD is not installed in your cluster"_ — the chart author is aware of the requirement.

[...]

PR #894, opened 2026-04-23
#893 - Fix deploy-k8s.sh to match project conventions

yoosful:
## Description

`deployments/scripts/deploy-k8s.sh` does not produce a working cluster from a clean state. Two independent failures surface before any workload runs:

1. **PodMonitor enabled by default** — `services.api.podMonitor` defaults to `true` in the sub-charts, so `helm install` fails with `no matches for kind "PodMonitor"` on any cluster without the Prometheus Operator. The umbrella quick-start chart and `docs/deployment_guide/appendix/deploy_minimal.rst` both disable this; the script does not.
2. **Backend-operator token uses removed `--service` flag and skips user creation** — The CLI removed `--service` in favor of `--user`, and tokens require a pre-existing user. The minimal deploy docs create a `backend-operator` user first and use `--user backend-operator`; the script still passes `--service` and silently produces an empty token, leaving all backend pods in CrashLoopBackOff.

This patch aligns `deploy-k8s.sh` with the minimal-deploy documentation:

- Append `podMonitor.enabled: false` to the generated `values-*.yaml` for both renders.
- Create the `backend-operator` user before token creation, and replace `--service` with `--user backend-operator`.

Verified end-to-end on a fresh EKS 1.30 cluster (sa-east-1): all core pods reach `1/1 Running`, backend-operator token is generated and consumed cleanly, no CrashLoopBackOff.

Issue #893

## Checklist
- [ ] New or existing tests cover these changes.

## Summary by CodeRabbit

* **Chores**
  * Updated Kubernetes deployment configuration to adjust pod monitoring settings for service sidecars and refined backend service authentication setup during environment initialization.

### DIFF
--- a/deployments/scripts/deploy-k8s.sh
+++ b/deployments/scripts/deploy-k8s.sh
@@ -413,6 +413,8 @@ sidecars:
     enabled: false
   oauth2Proxy:
     enabled: false
+podMonitor:
+  enabled: false
 EOF
 
     # UI values
@@ -509,6 +511,8 @@ services:
 sidecars:
   otel:
     enabled: false
+podMonitor:
+  enabled: false
 EOF
 
     log_success "Helm values files created"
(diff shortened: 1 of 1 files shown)


vvnpn-nv (approved):

vvnpn-nv:
thanks @yoosful lgtm

yoosful:
@vvnpn-nv Thanks, could you help me merge this? trying to add other fixes on top of this

PR #961, opened 2026-05-06
charts/backend-operator: configurable + relaxed startupProbe defaults

vvnpn-nv:
## Description

Two changes to backend-operator startup probes:

1. **Make the probe configurable.** Drop the hardcoded `failureThreshold` / `periodSeconds` / `timeoutSeconds` / `initialDelaySeconds` on `backendListener` + `backendWorker`; pull each through `services.{backendListener,backendWorker}.startupProbe.*` values. Same pattern as #5247b44f8 on the service-side chart.

2. **Relax the defaults.** `failureThreshold: 12 → 30`, `timeoutSeconds: 15 → 60` (others unchanged). The previous defaults assumed fast-bootstrap = 60s tolerance; managed K8s (AKS/EKS) cold-starts blow past that. New tolerance is 150s, harmless on fast clusters because the probe stops at first success — extra threshold/timeout has no overhead unless the service is actually broken. Only downside on a genuinely-broken pod is slower kubelet kill (~2.5 min vs ~60s); that's fine for a startup probe.

**Why.** Surfaced during full bring-up on Azure. On 2-CPU AKS system-pool VMs, `progress_check` (PyInstaller-bundled Python) cold-starts past the 15s `timeoutSeconds` on every invocation. Every probe times out, kubelet kills the container, the cycle repeats until helm `--wait` deadline fires — despite the listener/worker being functionally healthy and watching events the whole time.

**Verification.**
- `helm template` with no overrides → renders new defaults: 30 / 5 / 60 / 10
- `helm template --set services.backendListener.startupProbe.failureThreshold=12 ...` → overrides flow through correctly; backendWorker stays at defaults
- Live cluster: pod-level patch with same numbers (60 / 30) made backend-listener + backend-worker Ready in 2m20s on cold AKS nodes; they CrashLoopBackOff'd indefinitely with the previous defaults

Issue #None

## Checklist

## Summary by CodeRabbit

* **Chores**
  * Externalized startup probe configuration for backend services, enabling flexible tuning of health check parameters (failure thresholds, probe intervals, timeouts, and initial delays) through configurable values instead of hard-coded settings.

### DIFF
--- a/deployments/charts/backend-operator/values.yaml
+++ b/deployments/charts/backend-operator/values.yaml
@@ -335,6 +335,17 @@ services:
     ##
     apiBurst: 30
 
+    ## Startup probe tuning. Defaults sized for managed K8s (AKS/EKS) where
+    ## progress_check (PyInstaller-bundled Python) cold-starts past 15s on
+    ## small system-pool nodes. 60s timeout × 30 attempts = 150s tolerance.
+    ## Harmless on fast clusters (probe stops at first success).
+    ##
+    startupProbe:
+      failureThreshold: 30
+      periodSeconds: 5
+      timeoutSeconds: 60
+      initialDelaySeconds: 10
+
   ## Backend worker configuration
   ## Executes jobs received from the backend listener
   ##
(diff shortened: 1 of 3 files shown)


cypres (approved):

deployments/charts/backend-operator/values.yaml
```diff
@@ -335,6 +335,17 @@ services:
     ##
     apiBurst: 30
 
+    ## Startup probe tuning. Defaults sized for managed K8s (AKS/EKS) where
```
elookpotts-nvidia, line 338: Do we need to update the sibling README as well?

elookpotts-nvidia (approved):

PR #1030, opened 2026-05-20
BYO/Azure deploy: bash3 array, dead NGC block + MEK guard, GPU label, smoke-test failure exit code, single-pool auto-pick

vvnpn-nv:
## Description

Six fixes to make \`deploy-osmo-minimal.sh --provider byo\` work end-to-end on Azure AKS with a managed Postgres/Redis backend. All bugs hit on the same fresh-cluster bring-up; each commit is independently revertable.

Issue #None

### Commits

1. **\`storage/common.sh\`** — guard empty-array expansion under bash 3.2 (macOS default). \`\"\${addressing_style_arg[@]}\"\` errored \`unbound variable\` whenever the backend (azure-blob, gcs, swift) didn't pass an addressing_style. Replaced with the \`\${arr[@]+\"\${arr[@]}\"}\` idiom — no semantic change on bash 4+.

2. **\`deploy-k8s.sh\`** (three things, same file):
   - Remove the dead NGC pull-secret block in \`service_set_flags()\`. The workflow \`init-container\` + \`client\` images at \`\$OSMO_IMAGE_REGISTRY\` are public, so \`backend_images.credential\` is unused. Worse, the unconditional \`secretRefs[0]\` \`--set\` clobbered whatever the storage-values fragment put at index 0 (\`osmo-workflow-data-cred\`) — silently dropping it from the rendered pod spec. osmo-service then crashed with \`Failed to read secret directory /etc/osmo/secrets/osmo-workflow-data-cred\`.
   - Add \`db_has_existing_data()\` probe and use it to guard MEK regeneration. The existing logic preserved \`mek-config\` only when the K8s ConfigMap was present. When the namespace gets deleted but the managed Postgres database persists (common BYO cloud pattern), the script silently minted a fresh MEK against existing encrypted data — orphaning every encrypted column. Result: osmo-service / backend-operator crashed with \`jwcrypto.jwk.InvalidJWKValue\` on \`service_auth.private_key\` read. The probe is fail-closed on probe failure (return 2) so a transient PG outage during deploy doesn't authorize a fresh MEK.
   - Fix GPU detection label selector. GPU Operator NFD labels GPU nodes with \`nvidia.com/gpu.present=true\` (dotted key, boolean value); the previous selector \`nvidia.com/gpu=present\` was the taint shape and matched zero nodes, so \`render_gpu_pool_values\` always skipped the fragment.

3. **\`values/gpu-pool.yaml\`** — fix nodeSelector to use the real label (\`nvidia.com/gpu.present=true\`) and add \`default_platform: gpu\`. Without the latter, default-platform workflow pods land on system nodes that lack the \`nvidia\` RuntimeClass (GPU Operator's container-toolkit DaemonSet only deploys on labeled GPU nodes), failing at sandbox creation with \`no runtime for nvidia is configured\`.

4. **\`deploy-osmo-minimal.sh\`** — \`verify.sh\` failure was swallowed into \`log_warning\`; the script then printed \"OSMO deploy complete\" and exited 0 even when verify-hello CrashLoopBackOff'd. Now exits non-zero on smoke-test failure; opt-out with \`OSMO_TOLERATE_VERIFY_FAILURE=1\` or skip the phase entirely with \`SKIP_VERIFY=1\`.

5. **\`cli/pool.py\`** — \`fetch_default_pool()\` previously required \`osmo profile set pool <name>\` before any pool-dependent command worked. On clusters with exactly one pool (common fresh-deploy state) it now auto-picks. \`list_pools()\` is wrapped in try/except so a transport error during the fallback doesn't replace the user-facing message with a stack trace. New tests in \`src/cli/tests/test_pool.py\`.

## Checklist
- [x] New or existing tests cover these changes (\`src/cli/tests/test_pool.py\` for the CLI fix; deploy-script changes have no unit-test home and were verified by the live BYO/Azure bring-up that produced these fixes).
- [x] The documentation is up to date with these changes (inline help text on smoke-test failure includes the opt-out env vars).

## Test plan

- [x] \`storage/common.sh\` fix: re-ran \`deploy-osmo-minimal.sh --provider byo --storage-backend azure-blob --non-interactive\` on bash 3.2.57 (macOS) — previous \`unbound variable\` abort no longer reproduces.
- [x] Dead NGC block removal: re-ran the deploy; pod spec now mounts all three workflow-cred volumes (data/log/app) instead of dropping data-cred.
- [x] MEK guard: simulated the dangerous case by deleting \`osmo-minimal\` namespace while Postgres persisted; script now aborts with the documented remediation message instead of minting a fresh MEK.
- [x] GPU label fix: \`render_gpu_pool_values\` now detects GPU nodes on AKS with stock GPU Operator NFD labels. \`gpu_toleration\` pod template's nodeSelector matches.
- [x] \`default_platform: gpu\` in gpu-pool.yaml: \`verify-hello\` workflow now schedules on GPU nodes (where the \`nvidia\` RuntimeClass exists) instead of failing with \`no runtime for nvidia is configured\` on system nodes.
- [x] Smoke-test-fail-exit: tested by injecting a failing verify-hello — script now exits 1.
- [x] CLI single-pool auto-pick: ran \`osmo workflow submit\` against a single-pool cluster without setting \`osmo profile set pool\` — submit succeeds; new \`test_pool.py\` covers the 6 cases.

## Summary by CodeRabbit

* **New Features**
  * Added a pre-deploy safety probe that blocks minting of cluster keys when the database appears populated or probe status is uncertain and required config is missing.
  * Default pool is auto-selected when exactly one pool exists in the backend.

* **Bug Fixes**
  * Corrected GPU node detection to use the updated node label format.
  * Fixed optional argument expansion in secret creation.
  * Simplified image-pull secret handling and related service credential wiring.

* **Tests**
  * Added unit tests covering pool selection and related error cases.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/OSMO/pull/1030?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/deployments/scripts/deploy-k8s.sh
+++ b/deployments/scripts/deploy-k8s.sh
@@ -138,6 +138,13 @@ RUN_KUBECTL_APPLY_STDIN=""
 RUN_HELM="helm"
 RUN_HELM_WITH_VALUES=""
 
+# Populated by create_database() from the in-pod psql probe. Used by
+# create_secrets() to refuse a fresh MEK against a populated DB. Values:
+#   ""        — probe didn't run (in-cluster DB mode, or BYO without PG creds)
+#   <int>     — number of user tables in public schema (0 = empty DB)
+#   "UNKNOWN" — probe ran but result couldn't be parsed (treat as unsafe)
+DB_TABLE_COUNT=""
+
 ###############################################################################
 # Parse Arguments
 ###############################################################################
@@ -350,6 +357,8 @@ spec:
           psql -h \$PGHOST -U \$PGUSER -d postgres -c 'CREATE DATABASE \"${POSTGRES_DB_NAME:-osmo}\";' 2>&1 || echo 'Database may already exist (this is OK)'
           echo 'Verifying database connection...'
           psql -h \$PGHOST -U \$PGUSER -d ${POSTGRES_DB_NAME:-osmo} -c 'SELECT 1 as connected;' && echo 'SUCCESS: Database ${POSTGRES_DB_NAME:-osmo} is ready!'
+          COUNT=\$(psql -h \$PGHOST -U \$PGUSER -d ${POSTGRES_DB_NAME:-osmo} -tAc \"SELECT count(*) FROM information_schema.tables WHERE table_schema='public';\")
+          echo \"OSMO_DB_TABLE_COUNT=\$COUNT\"
   restartPolicy: Never"
 
     log_info "Creating database initialization pod..."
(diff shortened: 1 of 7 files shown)


vvnpn-nv:
Addressed all three CodeRabbit findings in `6d7bd999d` with a refactor instead of in-place patches:

**Option 1 (per discussion with reviewer):** fold the table-count probe into the existing `create_database()` in-pod psql script. The pod now runs `CREATE DATABASE` + verify-connection + count-tables in one shot and emits `OSMO_DB_TABLE_COUNT=N` to stdout. `create_database()` parses that into a script-global `$DB_TABLE_COUNT`:
- `""` if probe didn't run (in-cluster DB mode, no PG creds — both safe first-install states)
- integer on success
- `"UNKNOWN"` on pod failure / timeout

`create_secrets()` now reads the var directly — no function call, no `set -e` interaction, no second pod spawn. Side effects:
- **#3 (set -e tri-state)** → moot: no function return code to capture.
- **#1 (apply-manifest failure)** → moot: no second `kubectl apply`. If the existing `create_database` pod fails, `DB_TABLE_COUNT` is set to `"UNKNOWN"` and the MEK check refuses to proceed.
- **#2 (RESET_MEK ignored when mek_exists=false)** → fixed: the gating logic is now `if RESET_MEK=true then warn+mint else check $DB_TABLE_COUNT`. RESET_MEK overrides the safety check regardless of `mek_exists`.
- Net -55 lines, one fewer pod spawn per deploy (~30-60s saved on cold starts).

Diff: https://github.com/NVIDIA/OSMO/pull/1030/commits/6d7bd999d

RyaliNvidia (approved):

cypres (approved):

PR #1039, opened 2026-05-23
fix(deploy): key GPU pool detection on the TF label, not NFD's

vvnpn-nv:
## Description

Workflow pods stay stuck in `SCHEDULING` on a freshly-deployed cluster because [`gpu-pool.yaml`](deployments/values/gpu-pool.yaml) and [`deploy-k8s.sh:render_gpu_pool_values`](deployments/scripts/deploy-k8s.sh) key on the NFD-set `nvidia.com/gpu.present=true`, which only appears after the NVIDIA GPU Operator finishes installing. `deploy-osmo-minimal.sh` installs KAI + GPU Operator + OSMO and then submits `verify-gpu` in sequence — but NFD's labelling is async and can lag, so the smoke test gets submitted before the GPU node carries the label and sits stuck (no node matches its nodeSelector).

Switch both the detection check and the workflow-pod `nodeSelector` to `nvidia.com/gpu=present`, which `terraform/{azure,aws}/example/` already puts on the GPU node pool at provisioning time. No NFD timing dependency.

| Label | Set by | Available when |
|---|---|---|
| `nvidia.com/gpu=present` | TF `node_labels` on the GPU pool | Node creation (immediate) |
| `nvidia.com/gpu.present=true` | NFD DaemonSet (shipped with GPU Operator) | After GPU Operator install + NFD reconcile |

The pod toleration (`key: nvidia.com/gpu, operator: Exists, effect: NoSchedule`) already matches the TF-set taint `nvidia.com/gpu=present:NoSchedule` and is unchanged.

Also rewrites the stale comment in [`aws/example/example.tf`](deployments/terraform/aws/example/example.tf) that still referenced the old `sku=gpu:NoSchedule` taint (the resource hasn't used that key in some time).

### BYO behaviour change

BYO clusters whose GPU nodes use a different label (e.g. AKS Console default of `accelerator=nvidia`) will now silently skip GPU pool layering. The previous NFD label would also have failed there if GPU Operator wasn't installed, so this isn't a strict regression — but the contract shifts slightly toward "OSMO TF-style labelling required." `gpu-pool.yaml` carries an inline note telling BYO operators to override `nodeSelector` via a values file or to label their nodes `nvidia.com/gpu=present`. A follow-up could add an explicit `log_warning` in `render_gpu_pool_values` when GPUs are advertised on the cluster but no node carries the expected label.

Issue #None

## Checklist

## Summary by CodeRabbit

* **Documentation**
  * Improved GPU node pool configuration documentation across AWS and Azure deployments

* **Chores**
  * Standardized GPU node detection and scheduling configuration across deployments for consistent behavior

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/OSMO/pull/1039?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/deployments/scripts/deploy-k8s.sh
+++ b/deployments/scripts/deploy-k8s.sh
@@ -875,7 +875,7 @@ extra_values_flags() {
 # was passed via NO_GPU=0 + force). Replaces the 6.2-era `osmo config update`
 # CLI dance — in 6.3 ConfigMap mode the pool definition lives in Helm values.
 #
-# To force-enable on a cluster without the gpu.present label, set
+# To force-enable on a cluster without the `nvidia.com/gpu=present` label, set
 # OSMO_GPU_POOL_ENABLED=true. To force-disable, set NO_GPU=1.
 render_gpu_pool_values() {
     GPU_POOL_VALUES_FILE=""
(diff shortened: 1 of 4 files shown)


vvnpn-nv:
Closing — wrong fix. The label choice isn't the bug; KAI scheduler is not applying the right node selector to pods. Re-investigating.

PR #1090, opened 2026-06-10
Fix retry pods failing to mount generated file secrets

vvnpn-nv:
## Description
Retry now recreates generated file Secret resources when a task is rescheduled. Previously the retry path rebuilt the Kubernetes pod from `TaskGroup.convert_to_pod_spec(...)` but discarded the returned generated `FileMount` objects, so the retry `CreateGroup` payload only contained the pod. If the original generated Secret was already gone, the retry pod failed with `FailedMount: secret ... not found` and eventually `FAILED_START_TIMEOUT`.

This change includes the generated file Secret resources before the retry pod in the retry `CreateGroup` payload. If the Secret still exists, backend create already treats `AlreadyExists` as harmless.

Issue - None
## Checklist

Verification:
- `bazel test //src/utils/job/tests:test_task_pure`
- `bazel test //src/utils/job:job-pylint //src/utils/job/tests:test_task_pure-pylint`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved job retry behavior so rescheduled tasks now include all generated Kubernetes resources (secrets and pod configs) ensuring more reliable automatic recovery of failed work.

* **Tests**
  * Added test coverage validating that task retry/reschedule flows assemble and submit the full set of Kubernetes configuration resources in the correct order and report progress as expected.

### DIFF
--- a/src/utils/job/jobs.py
+++ b/src/utils/job/jobs.py
@@ -1184,7 +1184,7 @@ class UpdateGroup(WorkflowJob):
         # Get create job
         pod_list = {new_task.name: kb_objects.construct_pod_name(
             self.workflow_uuid, new_task.task_uuid)}
-        pod, _, _ = group.convert_to_pod_spec(
+        pod, files, _ = group.convert_to_pod_spec(
             new_task,
             spec,
             self.workflow_uuid,
@@ -1199,13 +1199,16 @@ class UpdateGroup(WorkflowJob):
             skip_refresh_token=True,
         )
         k8s_factory.update_pod_k8s_resource(pod, group.group_uuid, pool, workflow_obj.priority)
+        labels = pod.get('metadata', {}).get('labels', {})
+        k8s_resources = [file.secret(labels) for file in files.values()]
+        k8s_resources.append(pod)
 
         create_job = CreateGroup(
             backend=spec.backend,
             group_name=group.name,
             workflow_id=self.workflow_id,
             workflow_uuid=self.workflow_uuid,
-            k8s_resources=[pod],
+            k8s_resources=k8s_resources,
             user=self.user)
 
         reschedule_job = RescheduleTask(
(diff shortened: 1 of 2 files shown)


xutongNV (dismissed):

src/utils/job/jobs.py
```diff
@@ -1214,20 +1199,22 @@ def _retry_task(
             skip_refresh_token=True,
         )
         k8s_factory.update_pod_k8s_resource(pod, group.group_uuid, pool, workflow_obj.priority)
+        k8s_resources = [file.secret(group._labels(self.user, self.workflow_uuid))
+                         for file in files.values()]
+        k8s_resources.append(pod)
```
cypres:
```suggestion
        labels = group._labels(self.user, self.workflow_uuid)
        k8s_resources = [*(file.secret(labels) for file in files.values()), pod]
```

src/utils/job/jobs.py
```diff
@@ -1214,20 +1199,22 @@ def _retry_task(
             skip_refresh_token=True,
         )
         k8s_factory.update_pod_k8s_resource(pod, group.group_uuid, pool, workflow_obj.priority)
+        labels = pod.get('metadata', {}).get('labels', {})
+        k8s_resources = [file.secret(labels) for file in files.values()]
+        k8s_resources.append(pod)
```
cypres, line 1204:
```suggestion
        pod_labels = pod.get('metadata', {}).get('labels', {})
        k8s_resources = [*(file.secret(pod_labels) for file in files.values()), pod]
```

vvnpn-nv: Thanks. I agree with the more specific `pod_labels` name, but I am going to keep the two-step list construction here. The explicit list comprehension plus `append(pod)` is easier to read and matches the existing style in this area better than the starred generator expression.

xutongNV (approved):

cypres (approved):

Issue #911, opened 2026-04-28
OSMO Datasets Deprecation


ethany-nv:
## Summary

The OSMO datasets API and CLI will be deprecated, along with `dataset:` references within the workflow spec. URL-based I/O (`url:`) replaces them.

adelbertc:
Done in OSMO 6.3.

PR #1128, opened 2026-06-23
Convert OETF dataset fixtures to task I/O

ethany-nv:
## Description
Convert the remaining dataset-backed OETF workflow fixtures to supported workflow I/O patterns instead of deleting useful coverage.

This PR now:
- keeps the serial workflow coverage using task `{{output}}` / `{{input:N}}` dependencies, with no `downloadType` override in the fixture
- restores folder-input coverage as task-output folder fanout and selection
- keeps `osmo data` CLI surface coverage via the restored `dataset_cli.yaml` fixture, now running `osmo data` help checks instead of removed `osmo dataset` commands
- restores the corresponding OETF scenario wiring while keeping the heterogeneous multi-arch scenario internal-only
- removes stale OETF runner wording about dataset localpaths

The old `serial_workflow_update_dataset.yaml` fixture was removed because it existed specifically to exercise dataset update behavior. The old regex workflow was removed as well: URL inputs/outputs support `regex`, but this public OETF set does not have a portable storage URL fixture to exercise URL regex end-to-end, and the task-output rewrite did not test regex behavior.

Live validation was run against OSMO with `osmo workflow validate` and `osmo workflow submit` on `isaac-hil` using `platform=x86-5090`. These completed successfully:
- `serial-workflow-2882` (revised serial workflow without `downloadType`)
- `folder-input-test-12`
- `data-cli-2`

Issue #911

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a new folder-input workflow that validates selecting files from nested folders.
  * Updated workflow examples to use the newer data CLI flow and direct file-based task handoffs.

* **Bug Fixes**
  * Improved local path detection so nested non-file sections no longer interfere with file handling.
  * Fixed several workflow scenarios to read inputs from task outputs more reliably.

* **Documentation**
  * Refreshed test and CLI docs to match the updated data/storage workflow terminology.


test/workflow/folder_input.yaml
```diff
@@ -15,113 +15,118 @@ workflow:
...
-          ls -alR {{input:0}}
-
-          input_path={{input:0}}/{{workflow_name}}/folder_input_1
-          invalid_input_path={{input:0}}/{{workflow_name}}/folder_input_0
-
-          echo "input_path: $input_path"
-
-          # Check if the folder_input directory exists
-          if [ ! -d $input_path ]; then
+      platform: {{platform}}
```
jiaenren: This will probably fail in the isaac-nightly pool. Drop the `platform:` line from the resource block and the entry from default-values so that it uses the pool's default platform. Same below. See https://github.com/NVIDIA/OSMO/pull/1114.

test/scenarios/BUILD
```diff
@@ -83,22 +82,14 @@ oetf_scenario_test(
...
-# platforms or registry credentials needed).
-_KIND_RUNNABLE_SERIAL = ["test_serial_workflow_mounting"]
-
 [
     oetf_scenario_test(
         name = t.removeprefix("test_").replace("_", "-"),
         src = "serial.py",
         test_filter = "SerialWorkflows." + t,
-        tags = ["serial"] + (["kind"] if t in _KIND_RUNNABLE_SERIAL else []),
+        tags = ["serial"],
```
jiaenren: Can we also tag it with "kind" so that it runs in local kind deployment test?

jiaenren (approved):
Thanks, Ethan. Could you please run these test against staging env before merge?

RyaliNvidia (approved):

PR #1125, opened 2026-06-22
Add 6.4 dataset config cleanup migration

ethany-nv:
## Description
Depends on #1144.

Make the deployment upgrade path clean up deprecated live dataset configuration without restoring source-level dataset config support.

This PR adds a 6.4 pgroll data migration that deletes deprecated `DATASET` rows from the PostgreSQL `configs` table during upgrade. It preserves `config_history` rows so historical dataset config revisions remain available for read-only history flows handled by #1144.

ConfigMap-backed `dataset:` blocks are not rewritten by this migration. ConfigMap mode already ignores stale dataset blocks before loading configs, but admins should still remove those blocks from Helm values manually.

This PR also removes the stale `dataset:*` action from the shipped `osmo-user` default role policy so new service-chart installs no longer grant removed dataset permissions.

Validation:
- `git diff --stat FETCH_HEAD...HEAD` shows only deployment/migration files.
- `jq empty deployments/charts/service/migrations/005_v6_4_0_dataset_config_cleanup.json`
- `helm lint deployments/charts/service`
- `helm template test deployments/charts/service -f deployments/values/service.yaml`
- `POSTGRES_IMAGE=postgres:15.1 scripts/validation/test_dataset_config_cleanup_migration.sh`

Issue #911

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/OSMO/blob/main/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

## Summary by CodeRabbit

* **New Features**
  * Added an upgrade migration that automatically removes deprecated dataset configuration records from the database.

* **Bug Fixes**
  * Preserved existing audit/history entries and ConfigMap-based dataset settings during upgrades.
  * Removed the default dataset management permission from the standard user role.

* **Documentation**
  * Clarified upgrade behavior and noted that some legacy dataset settings in Helm values must be cleaned up manually.

PR #1144, opened 2026-06-25
Read legacy dataset config history

ethany-nv:
## Description
Allow legacy `DATASET` config history rows to remain readable without restoring live dataset config support.

This adds `DATASET` back as a history-readable config type, keeps dataset history data raw, and introduces operable config-history types/revision parsing for mutation paths. Rollback, delete-revision, tag, update, set, and current-config listing remain limited to supported live config types, so `DATASET` behaves like an invalid target for mutating operations.

Concrete occasions where this matters: `osmo config list`, direct `GET /api/configs/history` calls, and any UI/CLI flow that renders config-history revision lists from that endpoint. In practice this is only needed when legacy `config_history` dataset rows are intentionally preserved in the database instead of being deleted during upgrade.

The deployment migration/hook changes that preserve `config_history` belong in #1125 and are intentionally not included here.

Issue #911

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for viewing dataset entries in the config history list.
* **Bug Fixes**
  * Rollback, delete, and tag now validate and interpret config types more consistently using “operable” config types; rollback of dataset history revisions is blocked.
* **Documentation**
  * CLI help and subcommand option choices now consistently display only config types that can be operated on (including update, rollback, and tag).

### DIFF
--- a/src/utils/connectors/postgres.py
+++ b/src/utils/connectors/postgres.py
@@ -69,6 +69,20 @@ class ConfigType(enum.Enum):
 
 class ConfigHistoryType(enum.Enum):
     """ Type of configs supported by config history """
+    DATASET = 'DATASET'
+    SERVICE = 'SERVICE'
+    WORKFLOW = 'WORKFLOW'
+    BACKEND = 'BACKEND'
+    POOL = 'POOL'
+    POD_TEMPLATE = 'POD_TEMPLATE'
+    GROUP_TEMPLATE = 'GROUP_TEMPLATE'
+    RESOURCE_VALIDATION = 'RESOURCE_VALIDATION'
+    BACKEND_TEST = 'BACKEND_TEST'
+    ROLE = 'ROLE'
+
+
+class OperableConfigHistoryType(enum.Enum):
+    """Type of configs supported by config history mutations."""
     SERVICE = 'SERVICE'
     WORKFLOW = 'WORKFLOW'
     BACKEND = 'BACKEND'
--- a/src/utils/connectors/postgres.py
+++ b/src/utils/connectors/postgres.py
@@ -81,17 +81,21 @@ class ConfigHistoryType(enum.Enum):
     ROLE = 'ROLE'
 
 
+HISTORY_ONLY_CONFIG_HISTORY_TYPES = frozenset({
+    ConfigHistoryType.DATASET,
+})
+
 class OperableConfigHistoryType(enum.Enum):
     """Type of configs supported by config history mutations."""
-    SERVICE = 'SERVICE'
-    WORKFLOW = 'WORKFLOW'
-    BACKEND = 'BACKEND'
-    POOL = 'POOL'
-    POD_TEMPLATE = 'POD_TEMPLATE'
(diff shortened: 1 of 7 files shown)


cypres (dismissed):

RyaliNvidia (dismissed):

src/utils/connectors/postgres.py
```diff
@@ -69,6 +69,20 @@ class ConfigType(enum.Enum):
...
+    GROUP_TEMPLATE = 'GROUP_TEMPLATE'
+    RESOURCE_VALIDATION = 'RESOURCE_VALIDATION'
+    BACKEND_TEST = 'BACKEND_TEST'
+    ROLE = 'ROLE'
+
+
+class OperableConfigHistoryType(enum.Enum):
+    """Type of configs supported by config history mutations."""
     SERVICE = 'SERVICE'
     WORKFLOW = 'WORKFLOW'
```
elookpotts-nvidia: I'm worried about maintaining two enums here. `OperableConfigHistoryType` is `ConfigHistoryType` minus `DATASET`, so these lists can drift the next time a config-history type is added. If we need a distinct enum for Pydantic/OpenAPI validation, could we single-source the values with a shared tuple or construct the mutation-only enum from `ConfigHistoryType`?

elookpotts-nvidia (dismissed):

src/utils/connectors/postgres.py
ethany-nv: Good point. I tried constructing `OperableConfigHistoryType` dynamically from `ConfigHistoryType`, but Bazel mypy rejects functional enums whose member mapping is built from a comprehension (`Second argument of Enum() must be string, tuple, list or dict literal for mypy to determine Enum members`).

I pushed a type-checker-friendly version in `38fd4184`: `OperableConfigHistoryType` now sources each value from `ConfigHistoryType`, and there is a drift test that derives the expected operable values from `ConfigHistoryType - HISTORY_ONLY_CONFIG_HISTORY_TYPES`. That keeps Pydantic/OpenAPI behavior clean while making future drift fail tests.

Validated:
- `bazel test //src/service/core/config/tests:test_config_history_helpers //src/service/core/config/tests:test_config_history //src/cli/tests:test_cli`
- `bazel run //src/scripts:export_openapi > /tmp/osmo-1144-openapi.json`; `RollbackConfigRequest.config_type` references `OperableConfigHistoryType`, and that enum excludes `DATASET`.

ethany-nv: Agreed. I reworked this in 1cbf25ed so `ConfigHistoryType` is the single source of truth for config history values, and `OperableConfigHistoryType` is derived by subtracting the history-only set (`DATASET`). The rollback dispatch now compares against `ConfigHistoryType` values so mypy does not need to understand generated enum members. Validation: targeted Bazel tests passed, and regenerated OpenAPI still shows rollback using `OperableConfigHistoryType` without `DATASET` while read-history `ConfigHistoryType` still includes `DATASET`.

cypres (approved):

RyaliNvidia (approved):

ethanlookpotts (approved):

PR #1156, opened 2026-06-26
docs: build multiversion CLI pages from each version's own source (OSMO-6482)

samuelhanono-Nvidia:
## Description

The multiversion docs build rendered every version's CLI pages against `main`'s
parser, so release branches that still ship commands removed on `main` (e.g.
`release/6.3`'s `osmo bucket` / `osmo dataset`) failed with `NavigationException`.
Now each version's CLI is rendered from its own checked-out `src`, so releases
build without per-release freezing.

Changes:
- `docs/conf.py`: prepend the version being built (`SPHINX_MULTIVERSION_SOURCEDIR`,
  cwd fallback) to `sys.path` so `src.*` resolves to that release's source.
- `docs/locked_requirements.txt`: bump `sphinx-multiversion` to v0.2.7 (passes the
  env to the markdown pass too).
- `docs/Makefile`: add `--run-markdown-build` (markdown is opt-in in 0.2.7).
- `docs/_extensions/module_aliasing.py`: skip an `osmo.*` alias that can't import
  instead of failing the build.
- `.github/workflows/docs-deploy.yaml`: install `release/6.3`'s `src` reqs so its
  parser imports.

Validated: `make build-multiversion` builds `main` and `release/6.0`-`6.3`
(HTML + Markdown) with no `NavigationException`.

Issue #None

## Checklist

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -16,8 +16,7 @@
 
 # -- Path setup --------------------------------------------------------------
 
-import importlib
-import json
+import os
 import sys
 from pathlib import Path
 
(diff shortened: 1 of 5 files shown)


docs/conf.py
```diff
@@ -24,6 +26,51 @@
 # This is important for sphinx-multiversion which runs from temporary directories
 sys.path.insert(0, str(Path(__file__).parent.resolve()))
 
+
+def _multiversion_repo_root():
```
RyaliNvidia: I think you can perform the behavior you are trying to do by just doing something like:
```
# When sphinx-multiversion builds a release branch, the source files come from a
# temporary checkout while this config still comes from the current checkout.
# Prefer that versioned checkout for src.* imports so generated CLI docs match
# the release docs being rendered.
source_dir = Path(os.environ.get("SPHINX_MULTIVERSION_SOURCEDIR", Path.cwd())).resolve()
sys.path.insert(1, str(source_dir.parent))
```

SamuelHanono: That was my initial approach to this, but I found a problem:

sphinx-multiversion builds each version twice: once to make HTML (web pages), once to make Markdown. 

On the HTML run, SPHINX_MULTIVERSION_SOURCEDIR is set → the suggested code works → 6.3 uses 6.3's code.
On the Markdown run, that variable is missing → the code falls back to Path.cwd().

And Path.cwd() is not reliably the right version's folder. The tool sets the working directory to <checkout>/<path-where-you-ran-make>. So source_dir.parent only lands on the version's repo root if you happen to run make from inside docs/. If you run it as make -C docs … from the repo root, the math is off by one folder, import src falls back to main's code again → 6.3's Markdown build crashes with the exact NavigationException: ... bucket.

RyaliNvidia (dismissed):

RyaliNvidia (approved):

ethany-nv (approved):

PR #1162, opened 2026-07-01
[codex] Add Phase A self-hosted MCP authentication

SamuelHanono:
## Description

Adds Phase A authentication and delegation support for the self-hosted OSMO MCP endpoint.

This change:

- adds a stateless FastMCP Streamable HTTP service at `/mcp`, including health endpoints and an empty Phase A `tools/list` response;
- adds the MCP deployment, service, image targets, and network policy configuration to the service chart;
- publishes RFC 9728 protected-resource metadata and the corresponding `WWW-Authenticate` challenge;
- keeps `/mcp` protected by Envoy JWT validation and semantic authorization while bypassing the browser-oriented OAuth2 Proxy flow for MCP bearer tokens;
- validates browser `Origin` values against an explicit allowlist while continuing to support native clients that omit `Origin`;
- adds the `mcp:Access` RBAC action to the default OSMO roles; and
- adds focused MCP service and RBAC tests.

Deployment-specific OAuth resource URLs, identity-provider settings, public client IDs, and allowed origins remain environment configuration and are not hard-coded here.

Phase A intentionally exposes no operational MCP tools. It establishes the authenticated transport and authorization boundary that later phases will use.

Issue #None

## Validation

- `helm lint` passed with MCP enabled.
- `helm template` rendered the MCP workload and Envoy configuration successfully.
- `src/service/mcp/tests:test_server` passed both health and Streamable HTTP initialization/tool-list tests.
- End-to-end OAuth validation connected successfully and returned the expected empty tools list.
- Requests without a bearer token returned the expected `401` authentication challenge.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for a self-hosted MCP service, including deployment, networking, image build/load, and gateway routing.
  * Introduced MCP-specific access control and default permissions for standard user roles.

* **Bug Fixes**
  * Improved gateway handling for MCP traffic, including TLS, origin checks, and authorization behavior.
  * Updated build-local and local image workflows so MCP is included consistently during deployments and tests.

* **Documentation**
  * Expanded deployment and chart documentation to cover MCP configuration and TLS options.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -103,6 +103,7 @@ Entry point: `service/core/service.py`. Framework: FastAPI + Uvicorn + OpenTelem
 | `service/agent/` | Backend cluster integration via WebSocket. Receives node/pod/event/heartbeat streams from K8s clusters. |
 | `service/logger/` | Receives structured logs from osmo-ctrl containers. Persists task metrics to PostgreSQL. Distributed barriers via Redis. |
 | `service/delayed_job_monitor/` | Polls Redis for scheduled jobs, promotes to main queue when ready. |
+| `service/mcp/` | Serves the stateless Streamable HTTP MCP endpoint for external OSMO clients. |
 
 ### Python Libraries (`lib/`)
 
--- a/bzl/mypy/BUILD
+++ b/bzl/mypy/BUILD
@@ -37,6 +37,7 @@ mypy_cli(
         app_requirement("fastapi"),
         app_requirement("jinja2"),
         app_requirement("lark"),
+        app_requirement("mcp"),
         app_requirement("mypy-boto3-iam"),
         app_requirement("mypy-boto3-s3"),
         app_requirement("mypy-boto3-sts"),
--- a/deployments/charts/service/README.md
+++ b/deployments/charts/service/README.md
@@ -423,7 +423,7 @@ The chart does not create Redis credentials. Secret controllers such as External
 
 #### Gateway → Upstream TLS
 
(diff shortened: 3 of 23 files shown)


RyaliNvidia:
You need to update the oetf local images file

`--build-local` calls `image_specs()`, whose static list does not include MCP (`/test/oetf/local_images.py:60`). Therefore it never builds or loads:

//src/service/mcp:mcp_image_load_<arch>

So when the KIND variant tries to build OSMO and the MCP is enabled, the image won't be found

RyaliNvidia:
Another thing you need to add is tls support so the message between the MCP and envoy are encrypted. The messages between envoy and the OSMO services are already encrypted, but not between the new MCP and envoy

SamuelHanono:
> You need to update the oetf local images file
> 
> `--build-local` calls `image_specs()`, whose static list does not include MCP (`/test/oetf/local_images.py:60`). Therefore it never builds or loads:
> 
> //src/service/mcp:mcp_image_load_
> 
> So when the KIND variant tries to build OSMO and the MCP is enabled, the image won't be found

Add the MCP Bazel target and image tag to local_images.py.
Set its local KIND pull policy to IfNotPresent.
Update the OETF image-list and deployment tests.

SamuelHanono:
> Another thing you need to add is tls support so the message between the MCP and envoy are encrypted. The messages between envoy and the OSMO services are already encrypted, but not between the new MCP and envoy

Make MCP use the same existing TLS system as other OSMO services.
Generate a temporary self-signed certificate by default—no manual certificate required.
Allow an optional Kubernetes TLS secret.
Make Envoy connect to MCP using TLS 1.2–1.3.
Change MCP health probes to HTTPS when TLS is enabled.
Test TLS enabled, disabled, and certificate-secret modes.

RyaliNvidia (dismissed):

RyaliNvidia (approved):

elookpotts-nvidia (approved):

PR #1193, opened 2026-07-13
testbot: use picker metadata for PR targets

jiaenren:
## Summary

- Use Stage 2 picker metadata as the authoritative source for PR targets and rationales.
- Reserve filename inference for ad-hoc runs without metadata, and fail when workflow-supplied metadata is missing or invalid.
- Preserve selected-target order, disambiguate basename collisions, and label the PR section as "Targets selected".
- Exclude testbot-only changes from the Local KIND Deployment workflow.

## Root cause

In [PR #1184](https://github.com/NVIDIA/OSMO/pull/1184), the generator organized tests for `src/utils/job/task.py` into a new `test_task_units.py` suite. `create_pr.py` then tried to reverse that generated filename into a source path, inferred a nonexistent `task_units.py`, and therefore could not match the picker metadata containing the real target and its reason.

Generated test filenames are output artifacts, so they cannot reliably encode source identity: one source may have several suites, a suite may cover multiple sources, and semantic suffixes such as `_units` are legitimate.

Issue - None

## Impact

Testbot PR titles, commit subjects, selected-target lists, and target-picking rationales now remain correct for arbitrary test filenames, multiple targets, BUILD wiring changes, and malformed metadata. Future testbot-only changes no longer consume Local KIND deployment capacity.

## Validation

- `bazel test //src/scripts/testbot/tests:test_create_pr`
- `bazel test //src/scripts/testbot/tests:test_create_pr-pylint`
- `bazel test //src/scripts/testbot:testbot-pylint`
- `bazel test //src/scripts/testbot/tests/...`

## Checklist

### DIFF
--- a/.github/workflows/oetf-kind.yaml
+++ b/.github/workflows/oetf-kind.yaml
@@ -14,6 +14,7 @@ on:
       - 'bzl/**'
       - 'MODULE.bazel'
       - 'src/**'
+      - '!src/scripts/testbot/**'
       - '.github/workflows/oetf-kind.yaml'
   workflow_dispatch: {}
 
--- a/src/scripts/testbot/create_pr.py
+++ b/src/scripts/testbot/create_pr.py
@@ -12,6 +12,7 @@ Usage:
 """
 
 import argparse
+from collections import Counter
 import datetime
 import json
 import logging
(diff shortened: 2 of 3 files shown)


jiaenren:
Superseded by #1200, which contains the same commits under the corrected jiaenr/fix-testbot-target-mapping branch.

Issue #1212, opened 2026-07-20
Rsync uploads lose path semantics and can report false success


agirault:
## Summary

`osmo workflow rsync upload` currently has three related problems in its local rsync client path:

1. The bundled `gokr-rsync` v0.3.1 can encounter an internal transfer error, send zero files, and
   exit successfully.
2. OSMO normalizes local paths with `os.path.realpath()`, which removes a meaningful trailing `/`.
3. With Landlock enabled, the bundled client cannot upload an individual file or a named directory
   without a trailing `/`.

The underlying Landlock directory problem is tracked in
[gokrazy/rsync#66](https://github.com/gokrazy/rsync/issues/66).

## Expected path semantics

Given a local directory `/tmp/payload` containing `run.sh`:

```sh
# Copy the named directory.
osmo workflow rsync upload WORKFLOW TASK \
  /tmp/payload:/osmo/run/workspace/
# Expected: /osmo/run/workspace/payload/run.sh

# Copy only the directory contents.
osmo workflow rsync upload WORKFLOW TASK \
  /tmp/payload/:/osmo/run/workspace/
# Expected: /osmo/run/workspace/run.sh

# Copy an individual file.
osmo workflow rsync upload WORKFLOW TASK \
  /tmp/payload/run.sh:/osmo/run/workspace/
# Expected: /osmo/run/workspace/run.sh
```

These are the conventional rsync distinctions between a directory source with and without a
trailing slash.

### 1. Transfer failures can return success

The bundled `gokr-rsync` v0.3.1 can log an `OpenRoot(...): permission denied` error and
`received 0 names`, but exit with status 0. OSMO consequently reports that the upload completed
successfully even though no files landed.

This false-success behavior is fixed by
[`gokrazy/rsync@972b294`](https://github.com/gokrazy/rsync/commit/972b29462e87103c0dcb4fd58376fc1490ef4993),
which reports module `OpenRoot` errors as fatal. The fix is included in `gokr-rsync` v0.3.4, which
returns a nonzero status for the same failure.

### 2. OSMO removes the trailing slash

`validate_local_path()` resolves the source through `paths.resolve_local_path()`, which uses
`os.path.realpath()`. The resulting path no longer records whether the user supplied a trailing
slash.

OSMO therefore cannot preserve the intended distinction between copying a named directory and
copying its contents.

### 3. Landlock blocks files and named directories

The `gokr-rsync` client allowlists the source path with Landlock. For an individual file or a
[...]

PR #1213, opened 2026-07-21
#1212 - Fix rsync upload path semantics

agirault:
## Description

Upgrade bundled `gokrazy/rsync` from v0.3.1 to v0.3.4 so `OpenRoot` transfer failures propagate as nonzero exits. Preserve a local directory's trailing slash through path validation, and pass `--gokr.dont_restrict` only to the local upload client until gokrazy/rsync#66 is fixed. Downloads and the remote task-side server remain restricted.

Recheck the local source immediately before spawning rsync so a source removed after request parsing cannot become a successful empty upload. Exercise the real bundled binary for individual files, named directories, trailing-slash contents, and fatal missing-root behavior.

With v0.3.1, the real-binary regression suite reproduces both silent failures: a restricted named-directory transfer and a production-mode missing nested source each log an `OpenRoot` error and `received 0 names`, but exit zero. Both return nonzero with v0.3.4.

Testing:

- `bazel test //src/lib/rsync/tests:all --nocache_test_results --test_output=errors`
- `bazel build //src/runtime/cmd/rsync:rsync //src/lib/rsync:rsync_bin`

Issue #1212

Related: https://github.com/gokrazy/rsync/issues/66

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved upload reliability by verifying source files and directories still exist before transfers begin.
  * Preserved trailing slashes in directory paths to ensure contents are copied to the intended destination.
  * Updated transfer behavior for compatibility with the latest rsync runtime.
* **Tests**
  * Added coverage for missing sources, directory transfers, trailing-slash behavior, and upload/download command handling.


src/lib/rsync/rsync.py
```diff
@@ -616,8 +617,19 @@ async def upload(self) -> None:
...
+                )
+
             try:
-                rsync_args = [self._rsync_bin_path, RSYNC_FLAGS]
+                # TODO(gokrazy/rsync#66): Remove the client-side workaround once
+                # named directories and individual files work under Landlock.
+                rsync_args = [
+                    self._rsync_bin_path,
+                    RSYNC_FLAGS,
+                    RSYNC_DONT_RESTRICT_FLAG,
```
fernandol-nvidia, line 631: I am wondering if it makes sense to allow additional flags to be passed in from OSMO cli for `osmo workflow rsync <existing_params> -- <additional flags for rsync>`

So that users can add client side flags such as this without the need to clean anything up once the underlying rsync binary is fixed.

fernandol-nvidia (approved):

src/lib/rsync/rsync.py
agirault: That wouldn't hurt, but I'd advice doing it in a separate MR and keeping this default value since:
1. without this change, the default rsync behavior would not work, per test.
2. keeping this value to "not restrict" won't break the behavior even after/if this is fixed upstream.
3. will need an mr to upgrade the rsync binary eventually to pick up an upstream fix, so reverting this can be done at the same time.
4. exposing arbitrary additional flags is a larger scope.

If you feel strongly otherwise I'm happy to update, but this seemed like the lower hanging fruit focused on the issue.

fernandol-nvidia: Yea that makes sense, thanks for the clarification. LGTM

elookpotts-nvidia (approved):

PR #1194, opened 2026-07-14
Add workflow label support

jiaenren:
## Description

Issue - None

Tracking: OSMO-6501

### Purpose

OSMO needs immutable workflow metadata that survives admission, PostgreSQL persistence, restart/resubmit, and task-Pod creation. Existing mutable tags do not provide that contract or the Kubernetes labels needed for resource-attribution joins.

This draft integration PR adds generic workflow labels, with `PPP` as the first configured policy and reporting use case. Defaults remain inert: labels are optional and no key is required until an administrator configures a policy.

### Current integration contents

- Optional validated workflow `labels`, capped at 16 entries with reserved-key protection.
- Repeatable CLI/API label overrides and UI submit, resubmit, detail, and list support.
- ConfigMap-compatible per-key `off`, `warn`, and `enforce` policy under `labels_config.policy[]`.
- Warn-mode violations recomputed from the current policy for every workflow status, including `COMPLETED`.
- PostgreSQL-canonical labels for persistence, restart, and resubmit.
- Task-Pod label propagation with workflow < Pod-template < system/platform precedence.
- Label selectors for presence, missing keys, globs, and grouped alternatives.
- Label dimensions in task metrics and prototype PPP resource reporting.
- A retained OETF receipt workflow whose logs contain every full expected value, actual value, and PASS/FAIL result.
- Nullable `workflows.labels JSONB`, a generic `jsonb_ops` GIN index, and schema coverage across pgroll, in-code bootstrap, and the test database.
- Sparse-history pgroll reconciliation that treats data migrations explicitly and fails closed when trusted history disagrees with physical schema.
- Config PATCH semantics where ordinary lists replace prior values and explicit `$index` entries retain strategic merge behavior.

### Split review stack

PR #1194 remains the full integration/prototype branch. Reviewable slices are based on their real prerequisites, not on this monolith:

| Slice | PR | Head | Scope and state |
|---|---|---|---|
| B1 | #1216 | `23640ca5e` | Label validation library; checks passed |
| A1 | #1217 | `b6b0012fd` | Plain-list config replacement plus upgrade guidance |
| A3 | #1218 | `c3d3e9af9` | Safe pgroll reconciliation and released migration 005 |
| B2 | #1219 | `4d03ca4e7` | Label JSONB/GIN schema, stacked on #1218 |
| B3 | #1220 | `b8c57fda6` | Label policy config model, stacked on #1217 with B1 temporarily included |

[...]

PR #1218, opened 2026-07-21
fix(migrations): reconcile sparse pgroll history

jiaenren:
## Description

Issue - None

### Summary

- Add the released 005 v6.3 schema migration to the chart.
- Make pgroll initialization, status, completion, baseline, migration, and view-refresh failures fatal.
- Reconcile existing and sparse-history databases without re-running completed migrations.
- Fail closed when migration 005 is recorded or baselined but its resource columns are missing.
- Refresh versioned-schema views on every successful run.
- Package only migration JSON files and the runner in the Helm hook ConfigMap.

### Design context

This is prerequisite A3 for the OSMO-6501 workflow-label Track B stack. It deliberately contains no workflow-label schema or index behavior; B2 builds on it. The full v6.2 structural fingerprint is retained for safe reconciliation and remains an explicit review point while this PR is draft. The full prototype remains in #1194.

### Verification

- `bazel test //deployments/charts/service/migrations:test_run_migrations //deployments/charts/service/migrations:test_run_migrations-pylint --test_output=errors`
- `helm lint deployments/charts/service`
- `helm template test-release deployments/charts/service`
- Enabled-hook render contains exactly migrations 001 through 005 plus `run_migrations.sh`.
- `bash -n deployments/charts/service/migrations/run_migrations.sh`
- `git diff --check`

## Checklist

### DIFF
--- a/deployments/charts/service/migrations/005_v6_3_0_schema.json
+++ b/deployments/charts/service/migrations/005_v6_3_0_schema.json
@@ -1,0 +1,24 @@
+{
+  "operations": [
+    {
+      "add_column": {
+        "table": "resources",
+        "column": {
+          "name": "last_updated",
+          "type": "TIMESTAMP",
+          "nullable": true
+        }
+      }
+    },
+    {
+      "add_column": {
+        "table": "resources",
+        "column": {
+          "name": "last_usage_updated",
+          "type": "TIMESTAMP",
+          "nullable": true
+        }
+      }
+    }
+  ]
+}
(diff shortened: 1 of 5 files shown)


jiaenren:
Superseded: the label feature (6.4) now uses minimal pgroll migrations (005/006_v6_4_0_workflow_labels*) applied by the existing migration runner, so the sparse-history reconciliation rewrite and the unused 005 operator columns are no longer needed. Dropping A3 per scope decision; the label schema ships in #1219.

PR #1227, opened 2026-07-22
Add workflow label UI support

jiaenren:
## Summary

Issue - None

[OSMO-6501] Track B9 of the workflow-labels PR split — the final feature PR of the stack.

- Workflow list: Labels column plus `label` / `no_label` filter chips (selector syntax with `*` and `(a|b)` passes through to the list API).
- Detail page: stored labels and the recomputed warn-mode policy warnings (shown for every status, including completed runs).
- Resubmit panel: carries the workflow's labels forward and allows per-run edits via the label editor.
- New-workflow submit page: labels come from the spec editor (no separate widget that could drift), with the config panel surfacing parsed labels.
- Shared `workflow-labels` formatting module; API types consumed through the adapter layer per the repo convention.

Stacked on #1226; functionally depends on #1223 and #1224. Extracted from prototype #1194.

## Testing

Verified against `main`'s Next.js 16.2.11 (#1230):

- `pnpm vitest run` — 46 files, 1017 tests pass (includes new actions, columns, search-fields, shim, and label-formatting tests)
- `pnpm exec tsc --noEmit`, `pnpm exec eslint src --max-warnings 0`
- `pnpm build` passes
- Playwright journeys updated (submit form, detail overview, filters, resubmit panel); run in CI

## Checklist


src/ui/src/features/workflows/detail/components/panel/ui/workflow/workflow-details.tsx
```diff
@@ -186,8 +188,37 @@ const StatusDisplay = memo(function StatusDisplay({
...
+      <div
+        className="flex items-start gap-2 rounded-md border border-amber-200 bg-amber-50 p-3 text-amber-900 dark:border-amber-700/50 dark:bg-amber-900/20 dark:text-amber-200"
+        role="status"
+      >
+        <TriangleAlert
+          className="mt-0.5 size-4 shrink-0"
+          aria-hidden="true"
+        />
+        <div className="min-w-0">
+          <p className="text-sm font-medium">Workflow label policy</p>
```
cypres: There are other possible warnings than label policies.

Based on the diff, I assume we didn't have warning support previously, so it's good we are getting that now, but displaying warnings should not only support label policy warnings.

src/ui/src/features/workflows/list/lib/actions.ts
```diff
@@ -271,7 +275,7 @@ export interface ResubmitParams {
   /** Target pool for execution */
   poolName: string;
   /** Execution priority */
-  priority: string;
+  priority: SubmitWorkflowApiPoolPoolNameWorkflowPostParams["priority"];
```
cypres: What a name `SubmitWorkflowApiPoolPoolNameWorkflowPostParams`

src/ui/src/features/workflows/detail/components/panel/ui/workflow/workflow-details.tsx
jiaenren: Good point — the `warnings` field is generic, so I dropped the 'Workflow label policy' sub-heading and renamed the region to "Workflow warnings". It now renders whatever warnings the API returns. Fixed in f863dbbc9.

src/ui/src/features/workflows/list/lib/actions.ts
jiaenren: Agreed — aliased it to `SubmitWorkflowParams` in the adapter (matching the existing `... as WorkflowListEntry` pattern) and used that in actions.ts. Fixed in f863dbbc9.

cypres (approved):

fernandol-nvidia (approved):

PR #1238, opened 2026-07-27
Add external MCP profile settings mutation

SamuelHanono:
## Description

Issue #None

Starts the Phase 3 external MCP mutation stack with the shared write foundation
and `osmo_set_profile`.

- Generalizes the existing bounded, one-shot mutation relay for fixed
  POST/PATCH/DELETE routes, JSON object or string request bodies, and
  null/object/string JSON success values.
- Adds `osmo_set_profile` for the external CLI/Core-supported default-pool and
  email/Slack notification settings.
- Requires strict arguments, an exact JSON-null Core response, compact closed
  output, suppressed free-form write errors, and unknown-outcome guidance
  without retries.
- Updates the catalog and deployment documentation from 18 to 19 tools.

Profile setting support is limited to the default-pool and email/Slack
notification settings exposed by the CLI/Core contract.

### Stack

- Phase 3 stack PR 1/4.
- Targets `feature/mcp` after #1237 merged.
- #1239 is stacked on this PR.
- Nothing in this stack targets `main`.

### Validation

- `bazel --output_user_root=/tmp/osmo-bazel test --test_output=errors //src/service/mcp/tests:test_gateway //src/service/mcp/tests:test_tool_support //src/service/mcp/tests:test_profile //src/service/mcp/tests:test_catalog`
- `bazel --output_user_root=/tmp/osmo-bazel test --test_output=errors //test/smoke:mcp-checks-pylint`
- `bazel --output_user_root=/tmp/osmo-bazel build //src/service/mcp:mcp_binary //test/smoke:mcp-checks`
- `git diff --check`

The deployed smoke suite verifies discovery only for this new write tool; it
intentionally does not mutate a user's saved profile.

## Checklist

- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/OSMO/blob/main/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #1239, opened 2026-07-27
Add external MCP credential mutations

SamuelHanono:
## Summary

Adds the two external MCP credential mutations:

- `osmo_set_credential`
- `osmo_delete_credential`

This is Phase 3 PR 2/4. The prerequisite profile-setting support landed in
#1238, and this PR now targets `feature/mcp` directly. #1240 remains stacked on
this PR; nothing in this series targets `main`.

## Contract

- Reuses the canonical documented CLI/Core REGISTRY, DATA, and GENERIC request
  envelopes.
- Extracts the existing storage URI regex into a dependency-light public
  contract so DATA endpoint validation cannot drift between CLI/Core and MCP.
- Uses fixed credential routes, bounded string-only payloads, and one-shot
  writes with no automatic retry.
- Marks credential setting as open-world because Core may validate REGISTRY
  and DATA credentials against their external providers.
- Returns only credential name/type confirmation; payloads and legacy profile
  values are never projected.
- Rejects userinfo, query, and fragment components in registry/data profile
  fields before writing.
- Suppresses arbitrary upstream write details and reports an unknown outcome
  for ambiguous failures.
- Removes credential identifiers and payload data from telemetry.

The caller's MCP client owns the secret-bearing tool arguments it submits and
may retain them in its own transcript or logs. Core currently authorizes the
CLI's credential set POST route as `credentials:Create` and conflicts records
by profile, so some same-name updates are rejected rather than replaced. This
PR preserves that existing API/RBAC behavior.

## Validation

- `bazel --output_user_root=/tmp/osmo-bazel test --test_output=errors //src/cli/tests:test_credential //src/service/mcp/tests:test_credential_actions //src/service/mcp/tests:test_credentials //src/service/mcp/tests:test_catalog //src/service/mcp/tests:test_telemetry //test/smoke:mcp-checks-pylint`
- `bazel --output_user_root=/tmp/osmo-bazel build //src/service/mcp:mcp_binary //test/smoke:mcp-checks`
- `git diff --check`

Credential mutations remain manual Inspector checks against disposable
user-owned state; the automated deployment smoke test verifies discovery only.

Issue - None

## Checklist

### DIFF
--- a/src/service/mcp/tool_registry.py
+++ b/src/service/mcp/tool_registry.py
@@ -24,6 +24,7 @@ from mcp.types import ToolAnnotations
 
 from src.service.mcp import (
     apps,
+    credential_actions,
     credentials,
     health,
     pools,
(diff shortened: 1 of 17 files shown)


src/service/mcp/tool_registry.py
```diff
@@ -247,6 +248,35 @@ class ToolSpec:
             'Profiles and credential payloads are never returned.'
         ),
     ),
+    ToolSpec(
+        function=credential_actions.osmo_set_credential,
```
RyaliNvidia, line 253: Are we setting this as a closed world operation? I believe set credential validates the credential against like the registry or storage provider so I think it should probably be considered open

SamuelHanono: Yes that seems reasonable to me. Changing to open.

RyaliNvidia (approved):

jiaenren (approved):

PR #1241, opened 2026-07-27
Add external MCP app submission

SamuelHanono:
## Description

Adds `osmo_submit_app` and completes the external MCP's planned 26-tool catalog.

This is Phase 3 stack PR 4/4. It targets `feature/mcp`; nothing in this series targets `main`.

Issue - None

## Contract

- Validates all caller inputs before relay and resolves an explicit pool or the caller's accessible default/sole pool.
- Reads bounded app metadata, requires an exact matching name and UUID, and pins either the requested READY version or the newest READY version from at most 200 entries.
- Retrieves that concrete version's complete spec under a 128-KiB ceiling; it never submits a truncated spec.
- Reuses one shared workflow-submission path for pool encoding, template markers, overrides, priority, the fixed Core POST, response projection, and unknown-outcome handling.
- Sends only `app_uuid`, `app_version`, and non-default `priority` as fixed query options, then returns only workflow ID, app name/version, pool, priority, and confirmation.
- Does not expose local paths, environment injection, dry-run, rsync, or local-file expansion.

App metadata/spec reads require `app:Read`; an omitted pool additionally requires `profile:Read`; the final pool-scoped POST requires `workflow:Create`. Submission consumes compute, can leave a `FAILED_SUBMISSION` record, and is never automatically retried.

App specs and overrides are never returned or logged by MCP. Overrides are marked write-only, and callers are warned that their MCP client may retain submitted arguments.

## Validation

- `bazel --output_user_root=/tmp/osmo-bazel test --test_output=errors //src/service/mcp/... //test/smoke:mcp-checks-pylint`
- `bazel --output_user_root=/tmp/osmo-bazel test --test_output=errors //src/cli/tests:test_app`
- `bazel --output_user_root=/tmp/osmo-bazel build //src/service/mcp:mcp_binary //test/smoke:mcp-checks`

The deployment smoke test verifies the exact catalog but does not call `osmo_submit_app`; live submission remains a manual Inspector check against disposable user-owned state.

## Checklist

### DIFF
--- a/src/service/mcp/workflow_submission.py
+++ b/src/service/mcp/workflow_submission.py
@@ -37,7 +37,7 @@ from src.service.mcp.workflow_models import (
 _MAX_SUBMISSION_SPEC_BYTES = 128 * 1024
 _MAX_JSON_RESPONSE_BYTES = 64 * 1024
 _MAX_OVERRIDE_BYTES = 2048
-_CLI_TEMPLATE_MARKERS = ('{%%', '{{', '{#', 'default-values')
+_TEMPLATE_MARKERS = ('{%', '{{', '{#', 'default-values')
 
 
 def validate_pool_name(pool: str | None) -> str | None:
@@ -194,5 +194,5 @@ async def request_submission(
 
 
 def _is_templated_workflow(workflow_spec: str) -> bool:
-    """Match the CLI's lightweight template detection without its runtime."""
-    return any(marker in workflow_spec for marker in _CLI_TEMPLATE_MARKERS)
+    """Detect standard Jinja and OSMO workflow template markers."""
+    return any(marker in workflow_spec for marker in _TEMPLATE_MARKERS)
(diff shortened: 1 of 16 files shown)


src/service/mcp/workflow_submission.py
```diff
@@ -0,0 +1,198 @@
...
+from src.service.mcp.workflow_models import (
+    WORKFLOW_PRIORITIES,
+    WorkflowPriority,
+)
+
+
+_MAX_SUBMISSION_SPEC_BYTES = 128 * 1024
+_MAX_JSON_RESPONSE_BYTES = 64 * 1024
+_MAX_OVERRIDE_BYTES = 2048
+_CLI_TEMPLATE_MARKERS = ('{%%', '{{', '{#', 'default-values')
```
RyaliNvidia: Is the double % correct here?

elookpotts-nvidia (dismissed):

src/service/mcp/workflow_submission.py
SamuelHanono: I took it from here https://github.com/NVIDIA/OSMO/blob/main/src/cli/workflow.py#L590
We see the %%.  Should this also be just %?

RyaliNvidia (approved):

jiaenren (approved):

PR #1254, opened 2026-07-29
test(oetf): KIND-safe workflow-labels scenario (D1)

jiaenren:
## Summary

Issue - None

Adds an OETF end-to-end scenario for the workflow-labels feature, run in the KIND gate. It exercises the feature purely through the OSMO API (a sandboxed OETF test has no in-cluster kubeconfig):

- **Policy gate** off / warn / enforce via **validation-only** submits (no rows, nothing schedules): `warn` surfaces `warnings`, `enforce` rejects with 400 and leaves no row, `off` accepts anything.
- **Label-syntax rejection**: nested value, invalid key, empty value.
- **List filtering**: exact `label`, glob, and `no_label`.
- **Label round-trip**: a labeled workflow's labels survive submit → persistence → workflow API + list filtering.
- **Pod-label prefix** (`pod_label_prefix`): the prefix is validated against the merged key at submission (a bare key merges to a valid Kubernetes key; a key that already carries a prefix is rejected), and never leaks into the workflow API/list, which keep the bare keys.

Tagged `kind`, so it runs in the `oetf:deploy_and_run --env kind` gate (which `--build-local`s the stack and deploys it to KIND in DB mode). Serial + exclusive because the policy tests mutate `labels_config`, restoring the baseline in `tearDown`; on a ConfigMap-mode target those mutation tests `skipTest`, so the scenario is also portable to non-KIND deployments.

## Scope note (pod-object verification)

A literal in-cluster pod-label assertion is **not** reachable from the gate: OETF scenarios run via `bazel test` with only `OETF_URL` + auth + `OETF_POOL` forwarded, so the test can't reach the KIND cluster with kubectl. The KIND gate also doesn't run workflows to completion, so the end-to-end test asserts the **label round-trip** (submit → persisted labels echoed + filterable), not the run outcome. The actual pod stamping (`apply_workflow_labels` / the prefix application) is covered by unit tests; a true pod-object check would need kubeconfig forwarded into the OETF test env — a small runner change we can add later if wanted.

## Verification

- `bazel build //test/scenarios:workflow-labels` (and its pylint target) pass locally.
- The full-stack KIND gate (`oetf:deploy_and_run --env kind`) runs on this PR.

## Checklist

### DIFF
--- a/.github/workflows/oetf-kind.yaml
+++ b/.github/workflows/oetf-kind.yaml
@@ -167,7 +167,10 @@ jobs:
           # websocket-checks, ~5s) covers service+router+logger HTTP/WS;
           # templates + mount-validation (~6s) submit workflows so
           # backend-listener / backend-worker / osmo_ctrl runtime get
-          # touched.
+          # touched. workflow-labels exercises the label policy gate
+          # (off/warn/enforce) and runs one real labeled workflow to
+          # completion, covering the label spec field, admission, filters,
+          # and pod stamping path end to end.
           #
           # --jobs=1 serializes: chart's single-replica ingress-nginx 504s
           # under parallel scenario submits.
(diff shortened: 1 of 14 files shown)


jiaenren:
Superseded by #1260, re-created to base on #1259 (`pod_label_prefix`) so D1 is properly stacked. The base of this PR could not be changed via API because it was flagged as part of the now-merged B-stack. Same branch, same content plus the new pod_label_prefix coverage.

PR #1264, opened 2026-08-05
Preserve service auth identity in ConfigMap mode

jiaenren:
## Description

Issue #None

ConfigMap mode currently drops `service_auth` from the in-memory snapshot when the field is omitted from the ConfigMap. Every subsequent `ServiceConfig` construction then invokes the default factory and generates a new RSA signing identity. Tokens minted by one request can therefore fail validation against JWKS returned by the next request, which prevents backend listeners and workers from authenticating.

This change treats `service_auth` as a stable runtime identity while leaving ConfigMap data authoritative for all managed configuration:

- explicit ConfigMap/Secret-resolved `service_auth` takes precedence;
- cold starts recover the existing identity from PostgreSQL;
- reloads carry the identity forward from the previous snapshot;
- loads fail closed if no stable identity is available.

The change does not generate or persist new signing keys and does not hydrate any other ConfigMap-managed field from PostgreSQL.

## Validation

- `bazel test //src/service/core/config/tests:test_configmap_loader_unit --test_output=errors`
- Red/green regression coverage for cold-start hydration, reload carry-forward, explicit precedence, fail-closed behavior, and repeated key stability.
- ConfigMap integration and auth-service targets were attempted locally but could not start because this machine has no Docker socket; both use testcontainers and ran zero tests. The draft PR CI is expected to run them in the supported environment.

## Checklist

- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/OSMO/blob/main/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes; no user-facing documentation change is required.

## Summary by CodeRabbit

* **Bug Fixes**
  * Configuration loading now preserves service authentication across reloads.
  * Missing authentication can be recovered from persisted configuration when available.
  * Explicit authentication settings take precedence over recovered values.
  * Invalid configurations without a usable authentication source are rejected earlier.

* **Documentation**
  * Clarified configuration authority and service-authentication recovery behavior.

PR #1266, opened 2026-08-05
Preserve service auth identity in ConfigMap mode

jiaenren:
## Description

Issue - None

Supersedes #1264, which GitHub closed when its fork branch was renamed to comply with the `jiaenr/...` branch convention.

ConfigMap mode currently drops `service_auth` from the in-memory snapshot when the field is omitted from the ConfigMap. Every subsequent `ServiceConfig` construction then invokes the default factory and generates a new RSA signing identity. Tokens minted by one request can therefore fail validation against JWKS returned by the next request, which prevents backend listeners and workers from authenticating.

This change treats `service_auth` as a stable runtime identity while leaving ConfigMap data authoritative for all managed configuration:

- explicit ConfigMap/Secret-resolved `service_auth` takes precedence;
- cold starts recover the existing identity from PostgreSQL;
- reloads carry the identity forward from the previous snapshot;
- loads fail closed if no stable identity is available.

The change does not generate or persist new signing keys and does not hydrate any other ConfigMap-managed field from PostgreSQL.

## Validation

- `bazel test //src/service/core/config/tests:test_configmap_loader_unit --test_output=errors`
- GitHub `ci-public` passed on the identical commit in #1264.
- GitHub Local KIND Deployment passed on the identical commit in #1264.
- Live jiaenr ConfigMap-mode verification passed using images built from commit `964793040`:
  - ten JWKS requests returned the same persisted key ID, including after autoscaling;
  - the Isaac HIL listener connected all five streams;
  - the Isaac HIL worker connected and completed queue synchronization jobs;
  - both Argo applications reached Synced/Healthy.

Local testcontainer-backed integration targets could not start because this machine has no Docker socket; they ran in the supported GitHub CI environment instead.

## Checklist

- [x] No user-facing documentation change is required.

### DIFF
--- a/src/service/core/config/configmap_loader.py
+++ b/src/service/core/config/configmap_loader.py
@@ -131,10 +131,8 @@ class ConfigMapWatcher:
         backend_queue_updater: Callable[..., bool] | None = None,
         backend_test_updater: Callable[..., bool] | None = None,
     ):
-        # Kept for existing service/test wiring. ConfigMap loads must not
-        # hydrate ConfigMap-managed values from Postgres.
-        del postgres
         self._config_file_path = config_file_path
+        self._postgres = postgres
         self._event_recorder = event_recorder
         self._enable_reconciliation = enable_reconciliation
         self._backend_queue_updater = backend_queue_updater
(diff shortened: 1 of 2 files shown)


jiaenren:
Replaced by #1267, which uses the required in-repository NVIDIA/OSMO branch jiaenr/fix-configmap-service-auth-stability.

PR #1267, opened 2026-08-05
Keep service auth PostgreSQL-owned in ConfigMap mode

jiaenren:
## Description

Issue - None

ConfigMap mode currently drops `service_auth` from the in-memory snapshot when the field is omitted from the ConfigMap. Every subsequent `ServiceConfig` construction then invokes the default factory and generates a new RSA signing identity. Tokens minted by one request can therefore fail validation against JWKS returned by the next request, which prevents backend listeners and workers from authenticating.

This change keeps the JWT signing identity PostgreSQL-owned while leaving ConfigMap data authoritative for all other managed configuration:

- ConfigMap-supplied `service.service_auth` is discarded before secret resolution and cannot override the persisted identity;
- every new watcher hydrates the identity from PostgreSQL, even when an older global snapshot exists;
- reloads reuse only the watcher-local identity originally derived from PostgreSQL;
- loads fail closed if PostgreSQL does not provide a stable identity.

The change does not generate or persist new signing keys and does not hydrate any other ConfigMap-managed field from PostgreSQL.

## Validation

- `bazel test //src/service/core/config/tests:test_configmap_loader_unit --test_output=errors` (93 tests passed; target also ran mypy)
- Regression coverage verifies cold-start PostgreSQL hydration, ConfigMap override rejection before secret-file access, and DB-derived identity reuse across reloads.
- Live dev instance ConfigMap-mode verification using the initial fix confirmed stable JWKS keys, connected Isaac HIL listener/worker streams, successful queue synchronization, and healthy Argo applications.

Local testcontainer-backed integration targets could not start because this machine has no Docker socket; they run in the supported GitHub CI environment instead.

## Checklist

- [x] No user-facing documentation change is required.

## Bug Fixes
- Service authentication is now restored from persisted configuration when it is missing from a ConfigMap.
- Persisted authentication is retained across configuration reloads and takes precedence over ConfigMap-provided values.
- Configuration validation now requires service authentication and reports a permanent failure when no valid source is available.
- Configuration startup and reconciliation now consistently support persisted authentication.

## Documentation
- Startup behavior and authentication fallback handling have been clarified.

### DIFF
--- a/src/service/core/config/configmap_loader.py
+++ b/src/service/core/config/configmap_loader.py
@@ -131,10 +131,8 @@ class ConfigMapWatcher:
         backend_queue_updater: Callable[..., bool] | None = None,
         backend_test_updater: Callable[..., bool] | None = None,
     ):
-        # Kept for existing service/test wiring. ConfigMap loads must not
-        # hydrate ConfigMap-managed values from Postgres.
-        del postgres
         self._config_file_path = config_file_path
+        self._postgres = postgres
         self._event_recorder = event_recorder
         self._enable_reconciliation = enable_reconciliation
         self._backend_queue_updater = backend_queue_updater
@@ -244,6 +242,8 @@ class ConfigMapWatcher:
             if isinstance(section, dict):
                 _resolve_secret_file_references(section)
 
+        self._hydrate_service_auth(managed_configs)
+
         validation_errors = _validate_configmap_runtime_contract(managed_configs)
         validation_errors.extend(_validate_configs(managed_configs))
         if validation_errors:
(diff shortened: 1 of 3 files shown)


cypres (commented):
Why is "JWT signing identity" PostgreSQL-owned ?
Sounds like something we should move to a Kubernetes secret if it's secret, and config map if not secret

cypres (approved):
This fixes a problem related to changes in https://github.com/NVIDIA/OSMO/pull/1155

vvnpn-nv (approved):

PR #1270, opened 2026-08-06
Use Kubernetes Secrets for backend bootstrap authentication

vvnpn-nv:
## Description

Move backend bootstrap authentication from API-minted credentials to Kubernetes
Secrets while preserving the existing PostgreSQL access-token path.

Issue - None

## Summary

- Add API-server bootstrap authentication backed by Kubernetes Secret projections,
  so backend operators do not call the OSMO API to mint their initial credential.
- Resolve current and optional previous tokens to the fixed `osmo-backend` service
  identity while preserving PostgreSQL personal-access-token fallback.
- Add explicit service-chart credential sources:
  - `existingSecret.name` for production and multi-cluster deployments.
  - `managedSecret.name` for single-cluster development.
  - Deprecated `secretName` compatibility for existing values.
- Generate managed credentials inside Kubernetes with an ephemeral pre-install hook
  rather than in Helm templates.
- Keep KIND and private-AKS credential discovery and reconciliation from emitting,
  logging, or persisting token values in deployment-helper output.
- Update cloud installers, local/KIND flows, backend token refresh, documentation,
  and rotation guidance.

## Why

A backend operator needs a credential before it can register with the API server.
Having the API server generate that bootstrap credential creates a control-plane
dependency and was the regression discussed around #1155. This change moves only
backend bootstrap token handling to Kubernetes Secrets; the historical
database-backed `service_auth` concern remains separate.

## Managed Secret safety

- Helm renders only a short-lived ServiceAccount, Role, RoleBinding, and Job; it
  never renders the token or a Secret resource.
- The bootstrap command generates the token in memory and creates the Secret
  directly through the Kubernetes API without printing it.
- Temporary RBAC is removed after successful and failed hooks.
- Pre-upgrade validation preserves the live Secret and fails if it is missing
  instead of silently rotating it.
- The Secret is not Helm-owned, so rollback and same-name `managedSecret` to
  `existingSecret` transitions do not alter or delete it.
- Provenance labels and annotations identify chart-bootstrap Secrets without
  containing credential material.
- KIND checks return only a fixed presence marker. Private AKS performs token
  lookup, comparison, generation, and copying remotely and returns only fixed
  status markers through Azure Run Command logs.
- Managed mode is intentionally initial-install-only and intended for
  same-namespace development. Production remains externally managed.

## Authentication compatibility

- Mounted credentials are validated at API startup, compared in constant time,
  and rejected when malformed or duplicated.
- Rotation can expose current and previous tokens during a bounded overlap.
- Existing PostgreSQL-backed access tokens continue through the prior
  authentication path.
- Backend credentials always receive the reserved `osmo-backend` role and service
  identity.

## Verification

- `bazel test //src/service/core/auth/tests:test_backend_secret_auth //src/service/core/auth/tests:test_backend_secret_auth-pylint //src/service/core/auth/tests:test_backend_token_bootstrap //src/operator/utils/tests:test_login`
- `bazel test //run/tests:test_start_backend_kind //run/tests:test_start_backend_kind-pylint //deployments/scripts/tests:test_private_azure_backend_token`
- `bazel build //run:start_backend`
- `bash deployments/charts/service/tests/render-tests.sh`
- `helm lint deployments/charts/service`
- `helm lint deployments/charts/backend-operator`
- ShellCheck and shell syntax validation for the new deployment regression test
  and modified deployment script.
- `git diff --check origin/main...HEAD`
- Live deployment with two compute clusters using distinct `existingSecret`
  credentials against one control plane.
- Both backends registered successfully, reported nodes and platforms, and ran
  successful workflows on the Isaac and `osmo-backend` pools.

The checked-in Helm render matrix covers managed output with no Secret material,
existing mode, legacy compatibility, conflicting sources, multiple managed
credentials, upgrade guarding, and failure cleanup. Exercising the full
`managedSecret` install/upgrade/uninstall lifecycle on a live cluster remains a
recommended pre-merge validation.
## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added Kubernetes Secret-backed authentication for backend operators.
  * Helm deployments can create, reuse, rotate, and synchronize backend API token Secrets across clusters, including private Azure environments.
  * Local and minimal setups now use managed backend tokens and dedicated admin-password Secrets.

* **Documentation**
  * Updated deployment, authentication, upgrade, and troubleshooting guidance for Secret-based credentials and token rotation.

* **Bug Fixes**
  * Improved projected token-file rotation, credential validation, and deployment reliability.


vvnpn-nv:
Superseded by #1275 after republishing the source branch as `vpan/k8s-secret-backend-bootstrap`. The replacement points to the same commit (`f38094e27dd6e98651928df5bffd482aeaae3eab`).

PR #1289, opened 2026-08-11
testbot: target 3 files per run instead of 1

jiaenren:
## Description

Testbot PRs have been consistently small in scope — every recent one covers exactly **one** source file:

| PR | Files | Added | Target |
|----|-------|-------|--------|
| #1288 | 2 | +553 | copying.py |
| #1283 | 2 | +652 | downloading.py |
| #1280 | 2 | +634 | streaming.py |
| #1278 | 2 | +1193 | workflow.py |
| #1273 | 1 | +522 | common.py |

(The second changed file is the `BUILD` entry, not a second target.)

### Changes

- `max_targets` **1 → 3**, on both the dispatch input default and the schedule-path fallback. Nothing hardcodes the count — `SELECT_TARGETS_PROMPT.md` already renders `pick at most {max_targets}` and the generator prompt already loops per target — so this is a pure parameter change.
- `max_turns` **200 → 400**. The generator runs the full read/write/verify/coverage loop per target, and a single target has already exhausted the limit once (runs/26536045087, 101/100, after context auto-compaction).
- README table refreshed: it documented `max_turns=100` and `timeout_minutes=30` against the workflow's actual 200 and 60.

### Deliberately unchanged

- **`max_uncovered` stays 500.** It is a *per-target* cap (`_cap_ranges` is applied inside the per-file loop in `parse_coverage_entries`, with `remaining` reset per file), and recent targets carried only 35–231 uncovered lines. Raising it would change nothing; `max_targets` is the binding constraint. The per-PR ceiling still rises 500 → 1500 as a side effect of targeting 3 files.
- **Schedule and `timeout_minutes`.** Note the workflow is hourly with `concurrency.cancel-in-progress: true`, so a run is cancelled when the next hourly trigger fires — the effective wall-clock ceiling is ~60 min regardless of `timeout_minutes`. If 3-target runs start getting cancelled mid-generation, the fix is to slow the cron rather than raise the timeout.

### Verification

`bazel test //src/scripts/testbot:all //src/scripts/testbot/tests:all` — 19 targets pass, including `-pylint` siblings. Workflow YAML parses with the expected defaults.

Issue - None

## Checklist

## Summary by CodeRabbit

* **Chores**
  * Increased the default number of files processed per test generation run from 1 to 3.
  * Increased the default agent turn limit from 200 to 400.
  * Extended the default run timeout from 30 to 60 minutes.
  * Updated guidance to reflect the expanded processing limits.

### DIFF
--- a/.github/workflows/testbot.yaml
+++ b/.github/workflows/testbot.yaml
@@ -25,18 +25,18 @@ on:
     inputs:
       max_targets:
         description: 'Max files to target'
-        default: '1'
+        default: '3'
       max_uncovered:
         description: 'Max uncovered lines per target (0 = no cap)'
         default: '500'
       max_turns:
         description: 'Max Claude Code agent turns'
-        # Matches the schedule-trigger fallback below. Large picks
-        # (e.g., utils/job/jobs.py at 500 uncovered lines) need the
-        # extra headroom — runs/26536045087 hit max_turns at 101/100
-        # after Claude's 200K context window auto-compacted, wiping
-        # most of the exploration state mid-run.
-        default: '200'
+        # Matches the schedule-trigger fallback below. Scales with
+        # max_targets — the generator runs the full read/write/verify
+        # loop per target, and runs/26536045087 hit max_turns at 101/100
+        # on a single target after Claude's context auto-compacted,
+        # wiping most of the exploration state mid-run.
+        default: '400'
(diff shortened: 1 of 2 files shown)


jiaenren:
Superseded by #1290, which combines this with the other open testbot changes.

PR #1300, opened 2026-08-14
Use FastMCP OIDCProxy for MCP authentication

SamuelHanono:
## Summary

Upgrade the MCP service to FastMCP 3 and pass FastMCP's built-in `OIDCProxy` directly to the existing server as its `auth` provider.

FastMCP handles OAuth discovery, CIMD with DCR fallback, PKCE, token exchange, refresh, and encrypted state. The verified upstream Entra access token continues through the existing OSMO Gateway and RBAC path.

No standalone OAuth service or custom OAuth endpoints are introduced.

Issue #None

## Testing

- `bazel test //src/service/mcp/tests:all --test_output=errors --lockfile_mode=error` — 39/39 passed
- `bazel build //src/service/mcp:mcp_binary --lockfile_mode=error`

## Checklist

- [x] Tests cover the changes.
- [x] The stacked chart PR documents deployment and rollback.

## Summary by CodeRabbit

- **New Features**
  - Added optional OIDC/OAuth authentication for the MCP endpoint, including token verification, refresh, secure credential storage, and trusted redirects.
  - Added support for Streamable HTTP, stateless operation, and JSON responses.
  - Improved authenticated request handling and access-token relay.
  - Added deployment configuration for OAuth proxy routing, secrets, Redis, and configurable credentials.

- **Bug Fixes**
  - Sanitized validation and tool errors to avoid exposing sensitive details.
  - Improved middleware, tool discovery, and request metadata handling.

- **Documentation**
  - Expanded guidance for authentication modes, configuration, security, rollout, and troubleshooting.

### DIFF
--- a/src/service/mcp/auth.py
+++ b/src/service/mcp/auth.py
@@ -16,14 +16,12 @@ limitations under the License.
 SPDX-License-Identifier: Apache-2.0
 """
 
-import base64
 import dataclasses
-import hashlib
-import json
-from typing import Any, cast
+from typing import cast
 from urllib import parse
 
 from cryptography.fernet import Fernet
+from fastmcp.server.auth.jwt_issuer import derive_jwt_key
 from fastmcp.server.auth.oidc_proxy import OIDCProxy
 from fastmcp.server.auth.providers.jwt import JWTVerifier
 import httpx
@@ -42,7 +40,7 @@ class MCPAuthConfig(pydantic.BaseModel):
 
     model_config = pydantic.ConfigDict(hide_input_in_errors=True)
 
-    enabled: bool = pydantic.Field(
+    auth_enabled: bool = pydantic.Field(
         default=False,
         json_schema_extra={'env': 'OSMO_MCP_AUTH_ENABLED'},
     )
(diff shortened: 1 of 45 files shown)


cypres (dismissed):
LGTM

src/service/mcp/auth.py
```diff
@@ -0,0 +1,359 @@
...
+            'OIDC client secret',
+        ),
+        token_verifier=verifier,
+        base_url=cast(str, config.issuer_url),
+        resource_base_url=cast(str, config.issuer_url),
+        issuer_url=cast(str, config.issuer_url),
+        redirect_path='/auth/callback',
+        allowed_client_redirect_uris=config.allowed_client_redirect_uris,
+        client_storage=encrypted_store,
+        jwt_signing_key=signing_key,
```
cypres: https://gofastmcp.com/servers/auth/oidc-proxy#param-jwt-signing-key

> Default behavior (None): The key is deterministically derived from client_secret using HKDF, on every platform. Because the derivation is deterministic, the same key is produced across restarts as long as client_secret doesn’t change, so tokens remain valid without any extra configuration. This convenience makes it only suitable for development and local testing.


For our users I think it would be ideal to rely on the default behavior since it's means we don't have to worry about yet another secret to be handled.

RyaliNvidia (dismissed):

src/service/mcp/auth.py
SamuelHanono: his now uses FastMCP’s default HKDF-derived signing key from client_secret, and the separate signing JWKS/Vault wiring has been removed. Client-secret rotation will invalidate existing proxy sessions, so users will need to authenticate again.

RyaliNvidia (dismissed):

RyaliNvidia (approved):

jiaenren (approved):

PR #1307, opened 2026-08-17
Support Kubernetes Secret MEK rotation

vvnpn-nv:
Issue - None
Tracking - OSMO-6594

## What changed

- Read the MEK keyring exclusively from a typed Kubernetes Secret reference at a chart-owned mount path. Vault annotations, ConfigMap rendering, inline key material, and legacy MEK values are no longer supported.
- Load the keyring once at process startup, authenticate the persisted UEK and direct-MEK ciphertext inventory before readiness, and emit a sanitized rollout descriptor containing only key IDs, generation, and bundle digest.
- Add an opt-in, create-only bootstrap Job for disposable installs backed by a new empty database. The Job generates the Secret inside Kubernetes without placing key material in rendered manifests or release state and never overwrites or deletes an existing Secret.
- Add operator-driven `prepare`, `activate`, and `rewrap` lifecycle Jobs. Kubernetes Leases serialize attempts, and Pod ownership plus startup descriptors gate activation and rewrap.
- Support both managed mode, where the lifecycle Jobs create or patch the exact MEK Secret, and external mode, where the operator updates the Secret and the rewrap Job has exact-name read-only `get` access.
- Make rewrap explicit and restartable from the beginning. It compare-and-swap rewraps UEKs and registered direct-MEK configuration, then runs authenticated full inventories. Normal application reads no longer perform MEK rewrap writes.
- Keep every historical MEK in the Secret and reject removal. Rewrap completion is point-in-time evidence, not permission to retire an old key.
- Add unit, real-PostgreSQL, chart-render, deployment-script, and KIND-oriented rotation/bootstrap coverage.

## Design boundary

This implementation stores no MEK material or lifecycle metadata in PostgreSQL. It adds no MEK tables, triggers, functions, DDL, polling loops, or hot reloads. Coordination and rollout evidence are Kubernetes-native: a release-scoped Lease, immutable lifecycle Jobs, and sanitized startup descriptors from the complete Ready Pod cohort.

The database contains only the application ciphertext that already needs the MEK. UEK payloads and user plaintext do not change during rotation; only their encrypted wrappers and registered direct-MEK ciphertext are re-encrypted under the new current MEK.

## Validation

- `bazel test //src/utils/secret_manager/tests:test_secret_manager //src/utils/secret_manager/tests:test_mek_lifecycle //src/utils/connectors/tests:test_mek_reconciliation //src/utils/connectors/tests:test_mek_reconciliation_postgres //deployments/scripts/tests:test_mek_generation_guard`
- Maintained service-chart render suite
- Unified chart render/schema suite with pinned CNPG, Valkey, and RustFS dependencies
- `git diff --check` and shell syntax checks
- Independent security/correctness/lifecycle review approved

## Checklist

- [x] I am familiar with the Contributing Guidelines
- [x] New or existing tests cover these changes
- [x] The documentation is up to date with these changes

### DIFF
--- a/deployments/charts/osmo/README.md
+++ b/deployments/charts/osmo/README.md
@@ -195,7 +195,9 @@ secrets:
   objectStorage:
     existingSecret: osmo-object-storage
   masterEncryptionKey:
-    existingSecret: osmo-master-encryption-key
+    existingSecret:
+      name: osmo-master-encryption-key
+      key: mek.yaml
 ```
 
 Keep `embeddedDependencies.postgresql.enabled: false` (the default), then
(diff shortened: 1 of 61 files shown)


cypres (dismissed):

cypres (dismissed):

deployments/charts/service/README.md
```diff
@@ -451,7 +466,68 @@ The router was its own Helm chart prior to v6.3 and is now deployed as part of t
 | `services.router.startupProbe` | Startup probe configuration | See values.yaml |
 | `services.router.readinessProbe` | Readiness probe configuration | See values.yaml |
 
-The router reads the same `services.configFile.path` as the API service. When `services.configFile.enabled: false` (default), the router gets `--config <path>` as a CLI arg. The API service ignores `services.configFile.path` unless `services.configFile.enabled: true`, so setting just the path lets you point the router at a vault-injected config without affecting the API service.
+The router and all other control-plane database consumers read the MEK through
+the typed `services.masterEncryptionKey.existingSecret` reference. Production
+installs should create and manage that Secret outside Helm.
+
+For a disposable test install, set
```
ecolternv: This isn't meant to be a disposable test install. THis is meant to be production. The user should never need to supply the MEK in a production environment, we should take care of that for them

deployments/charts/service/README.md
```diff
@@ -451,7 +466,68 @@ The router was its own Helm chart prior to v6.3 and is now deployed as part of t
...
+`services.masterEncryptionKey.bootstrap.enabled=true`. A pre-install hook uses
+the configured kubectl image and a short-lived, namespace-scoped ServiceAccount
+to generate the initial 256-bit key inside Kubernetes and create the named
+Secret only when it is absent. MEK material is never rendered into Helm output
+or stored in Helm release state. The hook preserves an existing Secret, and a
+pre-upgrade hook fails if the Secret was deleted instead of silently generating
+a new key that cannot decrypt the database. The Secret persists after uninstall
+and must be deleted explicitly when the disposable environment is removed.
+
+To rotate the MEK, update that Secret in two separate Kubernetes writes. First,
```
ecolternv: We should automate this. We can have a cronjob as part of this helm chart that runs some "MEK rotate" job.

The cluster admin should really never need to think about the MEK.

deployments/charts/service/README.md
```diff
@@ -451,7 +466,68 @@ The router was its own Helm chart prior to v6.3 and is now deployed as part of t
...
+add the new JWK to `meks` while leaving `currentMek` unchanged (prepare). Use
+`kubectl logs` on every live consumer and wait for its `MEK consumer status`
+record to list the new key in `loaded_kids`; the same state is durably recorded
+by Pod UID in `public.mek_consumer_status`. Then change only `currentMek`
```
ecolternv: This seems pretty complicated, we've added like 5 new postgres tables just to track MEK rotation. Surely there is some simpler way that doesn't involve all of this extra state being tracked in the database.
```
1. key-rotation-job adds new MEK to  k8s secret
2. key-rotation-job waits for secret to be propagated to osmo processes (5 mintues? An hour? However long it takes for k8s secret projections to converge)
3. key-rotation-job sets the new MEK to be the active key
4. Again, key-rotation-job waits for convergence
5. key-rotation-job traverses all UEKs in the database and re-wraps them
6. key-rotation-job deletes the old MEK from the secret```

File: deployments/charts/osmo/README.md [hunk 4]
```diff
@@ -219,14 +221,74 @@ postgresql:
...
+diagnosis and is removed automatically after a successful retry.
+
+Rotate the Secret with two separate updates: add the new JWK while leaving `currentMek`
+unchanged, verify every live pod's `MEK consumer status` log lists the new key,
+then change only `currentMek`. Per-Pod-UID state is also stored in
+`public.mek_consumer_status`, and a pod that skipped the projected PREPARE
+revision can activate only an exact fingerprint already registered by another
+consumer. Existing UEK key material is preserved
+while its wrapper is moved to the active MEK; direct-MEK dynamic-config
+ciphertext is re-encrypted in bounded, durably checkpointed background batches.
```
ecolternv: >durably checkpointed background batches
Does this mean if we re-encrypt 100 UEKs and then the process crashes and restarts, we will not need to re process those 100 UEKs? I think its ok if we leave out hte checkpointing. There will be on the order of <1000 UEKs total in the DB, so if we need to reprocess them all, its not a big deal, and the complexity of this isn't worth it.

src/utils/connectors/postgres.py
```diff
@@ -1152,6 +1247,174 @@ def _init_tables(self):
...
+                first_seen_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
+            );
+        '''
+        self.execute_commit_command(create_cmd, ())
+        self.execute_commit_command('''
+            ALTER TABLE public.mek_key_registry
+            ADD COLUMN IF NOT EXISTS remaining_references INTEGER;
+        ''', ())
+
+        self.execute_commit_command('''
```
ecolternv: Why do we create this table and then run several command to alter the table we just created? Why not just do it all in one shot?

ecolternv (dismissed):

ecolternv (approved):

cypres (approved):

PR #1315, opened 2026-08-19
Close Redis log readers when clients disconnect

jiaenren:
## Description

Issue - None

Close Redis-backed log response iterators on every response teardown path, and propagate generator closure through the log filter and formatter to the Redis reader. Keep an active `http.disconnect` listener for quiet Redis streams, including when the server advertises ASGI 2.4, while translating only downstream send failures to `ClientDisconnect`; Redis/read failures remain operational errors.

File-backed completed-log responses are unchanged.

## Root cause

Starlette's ASGI 2.4 streaming path detects disconnects only when a body send fails. A quiet Redis `XREAD` stream may not emit another body chunk after the client leaves, so the request coroutine and its Redis client remain alive indefinitely. Closing the outer response alone was also insufficient because closure did not propagate through both wrapping async generators.

## Performance impact

The formatted-log hot loop is unchanged. The fix adds one disconnect-listener task per active Redis log stream and teardown-only `aclose()` calls; it adds no per-line network request, lock, copy, sleep, buffer, or formatting work.

## Validation

- `bazel test //src/service/core/workflow/tests:all //src/utils/connectors/tests:all //src/service/core/workflow:workflow-pylint //src/utils/connectors:connectors-pylint --test_tag_filters=-requires-network --test_output=errors --cache_test_results=no` — 15/15 passed
- OETF `logger-connectivity` against the patched KIND deployment — passed end to end
- Leak reproduction: baseline retained 20 active `XREAD` clients after 20 disconnected streams; the patched direct Uvicorn path retained 0 after 5 disconnected quiet streams
- Independent code review completed with no remaining Critical or Important findings

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved streaming log handling when a client disconnects, ensuring resources are cleaned up promptly.
  - Prevented disconnected log streams from leaving background readers, formatters, or Redis connections open.
  - Preserved underlying connection errors so failures are reported accurately.
  - Improved reliability for log streaming through middleware and during early client disconnects.
  - Ensured cleanup completes even when streaming tasks are cancelled.

### DIFF
--- a/src/service/core/workflow/BUILD
+++ b/src/service/core/workflow/BUILD
@@ -29,6 +29,7 @@ osmo_py_library(
     ],
     deps = [
         requirement("fastapi"),
+        requirement("starlette"),
         requirement("redis"),
         requirement("pyjwt"),
         requirement("kubernetes"),
--- a/src/service/core/workflow/tests/BUILD
+++ b/src/service/core/workflow/tests/BUILD
@@ -76,6 +76,7 @@ py_test(
         "//src/lib/utils:osmo_errors",
         "//src/service/core/workflow",
         "//src/utils/job",
+        osmo_requirement("starlette"),
     ],
 )
 
--- a/src/service/core/workflow/tests/test_workflow_service_units.py
+++ b/src/service/core/workflow/tests/test_workflow_service_units.py
@@ -21,7 +21,10 @@ import unittest
 from types import SimpleNamespace
 from unittest import mock
 
+from starlette.applications import Starlette
+from starlette.middleware.base import BaseHTTPMiddleware
 from starlette.requests import ClientDisconnect
+from starlette.routing import Route
 
 from src.lib.utils import osmo_errors
 from src.service.core.workflow import objects, workflow_service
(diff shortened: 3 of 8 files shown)


jiaenren:
Superseded by #1317, which is merged-ready and green.

#1317 takes a different approach to the same bug and additionally covers ground this PR did not:

- Instead of rewriting the ASGI scope to report `spec_version: '2.3'` so Starlette takes its pre-2.4 branch, it owns the disconnect race directly in an anyio task group, so behaviour no longer depends on which branch Starlette happens to take.
- It carries this PR's `Shield Redis reader cleanup from cancellation` fix, which turned out to be essential — verified in KIND, the response-layer fix alone still leaked 2 XREAD clients per 20 disconnects, and only the shield brought it to 0.
- It converts all six streaming endpoints (workflow logs both Redis and file backed, workflow spec, app spec, and the router's two webserver proxy responses) rather than the one, and replaces `check_client_version` with pure ASGI middleware so `BaseHTTPMiddleware` stops swallowing the disconnect signal for every streaming endpoint.

Thanks — the shield commit here was the key insight and is preserved in #1317.

PR #1316, opened 2026-08-20
Ecolter/self contained install

ecolternv:
## Description
Add backend chart into OSMO umbrella chart.
Allow full self-contained install

Issue - None

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added unified Helm profiles for control-plane, compute-only, and self-contained development environments.
  * Added configurable compute services, monitoring, networking, RBAC, priority classes, and test-runner resources.
  * Added automatic backend-token and master-encryption-key bootstrapping and validation.
  * Added external control-plane connectivity and token-authenticated compute deployments.

* **Bug Fixes**
  * Deployments now skip backend test synchronization without a configured test-runner namespace.
  * Control-plane resources are omitted when the control plane is disabled.

* **Documentation**
  * Expanded installation, configuration, production, profile, and legacy deployment guidance.


deployments/charts/osmo/profiles/split-plane-compute.yaml
```diff
@@ -0,0 +1,33 @@
...
+    enabled: false
+  valkey:
+    enabled: false
+  objectStorage:
+    enabled: false
+
+externalUrl: https://osmo.example.com
+
+compute:
+  backendName: default
```
vvnpn-nv: will this cause multiple backends to be called `default` if someone deploys the split plane with multiple backends attached?

ecolternv: Yes. I added some additional documentation in the README.md to address this. If you have multiple backends, then you override this.

```
helm install ... --values split-plane-compute.yaml --set-string compute.backendName='my-backend'
```

tdewanNvidia (dismissed):

deployments/charts/osmo/profiles/split-plane-compute.yaml
vvnpn-nv: Can we make backendName empty/required whenever the compute plane is enabled, and only set default explicitly only in `kind-self-contained.yaml`? Since helm cannot verify uniqueness across clusters, any copied installation will still validly install but can silently collide on the name default

vvnpn-nv (dismissed):

vvnpn-nv (approved):

tdewanNvidia (approved):

PR #1318, opened 2026-08-20
Add KIND regression for Redis reader cleanup

jiaenren:
## Description

Issue #None

Add a KIND-only regression scenario for the Redis log-reader cleanup fixed by #1315.

The scenario submits a quiet logger workflow, opens and forcibly disconnects 20 direct live-log clients, and verifies that the service returns to its baseline number of Redis `XREAD` clients. It is marked exclusive because it measures a shared Redis client count. The scenario macro now supports inheriting `HOME` and `KUBECONFIG` so the test can perform read-only KIND inspection.

This PR is stacked on #1315 and should be merged after the base fix.

Validation:

- `bazel run //test/oetf:run -- --env kind --tags redis` — 1/1 passed in 55.93s
- Without the owner-level cancellation shield from #1315, the regression retained 4 additional Redis `XREAD` clients after 20 disconnects.

## Checklist

### DIFF
--- a/test/oetf/bzl/oetf.bzl
+++ b/test/oetf/bzl/oetf.bzl
@@ -67,6 +67,7 @@ def oetf_scenario_test(
         workflow_data = ["@osmo_workspace//test/workflow:all_workflow_yamls"],
         tags = [],
         test_filter = None,
+        env_inherit = [],
         size = "large",
         timeout = "long"):
     """A scenario test — submits a workflow and asserts outcome.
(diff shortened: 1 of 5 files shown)


jiaenren:
Closing along with #1315, which this was stacked on. The fix itself now lives in #1317.

This scenario was genuinely valuable — it is what caught that the response-layer fix alone still leaked (2 XREAD clients per 20 disconnects), and it is how the final fix was verified at 0.

It is **not** carried into #1317, deliberately. Adding it there was tried and reverted: the KIND job takes an explicit `--target-pattern` allowlist of three *negative* scenarios that never run a workload, and on a GitHub runner this one fails with `FAILED_IMAGE_PULL` (no Docker Hub access for `python:3.10-slim`) plus `no nodes with enough resources: CPU cores`. `logger-connectivity` is excluded from that allowlist for the same reason.

So there is currently **no CI regression protection for this bug**. Closing the gap needs either image preloading in the KIND job or a scheduled dev-instance run.

The scenario is not lost — recover it any time with:

```
git fetch origin pull/1318/head:redis-reader-cleanup-oetf
```

PR #1320, opened 2026-08-20
Optimize username resolution with durable identities

ecolternv:
## Description

Username filters currently resolve identities with a large `UNION` over `users`, `workflows`, `apps`, and `app_versions`. This is necessary today because historical workflow and app owners are stored as text and may no longer have a corresponding row in `users`.

This change keeps that historical ownership model, but makes `users` authoritative for identities that can currently be filtered:

```text
users ──ON DELETE CASCADE──> profiles / credentials / roles / keys / tokens

workflow / app / app-version owner fields remain historical text
```

A migration backfills every existing historical owner into `users`, removes orphan credentials, adds the credential-to-user cascading foreign key, and creates an index for base-username lookup. The resolver then queries only `users` instead of scanning the historical tables.

Deleting a user removes account-scoped data while preserving their workflows and apps. The deleted identity is no longer filterable; recreating the same user immediately restores filtering of their historical records. Qualified names match exactly, while a base name resolves every matching current identity without duplicating results.

### Performance

On representative PostgreSQL data, base-name lookup improved from 40.626 ms and 836 shared-buffer hits to 0.070 ms and 3 hits. Exact lookup takes 0.058 ms and 3 hits. Neither plan scans workflow or app tables.

### Verification

- Migration, resolver, workflow/task/app filter, auth, task database, role-sync, and authz integration tests pass.
- Coverage includes exact, missing, ambiguous, multiple, duplicate, deletion, recreation, backfill, cascade, and indexed lookup cases.
- The migration and complete smoke suite passed on ecolter-dev.

Issue - None

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Consolidated user records into a single users table.
  * Improved username matching for exact names, base usernames, recreated users, and historical workflow data.
  * Added cascading credential cleanup when users are removed.
  * Added migration support for backfilling users and removing orphaned credentials.

* **Bug Fixes**
  * Prevented ambiguous or missing-user filters from returning incorrect results.
  * Preserved historical workflow, task, and app records after user deletion or recreation.

* **Tests**
  * Added comprehensive coverage for migrations, filtering, identity resolution, and database behavior.


cypres:
I would prefer keeping the users table, we can set a boolean flag when "deleting" a user. 

That is a bit of extra complexity in the application layer, but less than maintaining a user_identities table IMHO.

Alternatively we could set `ON DELETE SET NULL` on the FKs like workflows, to allow deleting a user but keeping their workflows.

and finally, we want to keep the user information on a workflow, but not delete the workflow, we can simply not have a foreign key on that table.

cypres (approved):

RyaliNvidia (approved):

PR #1328, opened 2026-08-26
Publish FastMCP's OAuth endpoints under the MCP path

jiaenren:
First of a stack simplifying MCP authentication. This PR narrows the gateway surface; later PRs remove the second auth mode and shrink the configuration.

Issue #None

## Problem

`auth.py` passed the gateway origin as all three of `base_url`, `resource_base_url` and `issuer_url`. FastMCP builds its **advertised** OAuth endpoint URLs from `base_url`, so every endpoint was advertised at the shared gateway root next to `/api` and `/cli` — and each one needed its own exact method/path Envoy route plus a duplicate `ext_authz` bypass matcher. Fourteen of each, listed twice in the template and a third time in CI.

## Change

Give the three parameters their three distinct meanings, derived from one value:

| Parameter | Value | Why |
|---|---|---|
| `base_url` | `https://host/mcp` | operational OAuth endpoints are advertised here |
| `resource_base_url` | `https://host` | RFC 9728 resource identity stays `/mcp`, not `/mcp/mcp` |
| `issuer_url` | `https://host/mcp` | what `authorization_servers` advertises, pointing clients at RFC 8414 path-aware discovery |

The MCP SDK registers its OAuth handlers at fixed root paths regardless of `base_url` (`mcp/server/auth/routes.py:92-114`), so the gateway publishes the `/mcp` prefix and rewrites it off before forwarding. The gateway now encodes FastMCP's **mount offset** rather than its **endpoint inventory** — strictly less coupling than the fourteen named routes it replaces.

The templated Lua Origin filter is replaced by FastMCP's `HostOriginGuardMiddleware`, configured from `services.mcp.allowedOrigins`.

## Result

```
Envoy template:                     1,372 → 1,269 lines   (wc -l)
MCP routes in the rendered config:     17 →     6         (grep -c '- name: mcp-')
```

Direct mode is untouched and still renders, so this is independently deployable.

## Two subtleties worth reviewer attention

- **The `/mcp/` prefix publishes the container's whole root namespace.** The health routes are registered at the app root, so they are explicitly carved out with a 404 route ahead of the prefix. Any future non-OAuth root route must be carved out too; there is a comment saying so.
- **The deployment origin is always added to the Origin allowlist.** Supplying any explicit allowlist disables FastMCP's same-origin fallback for non-loopback hosts (`fastmcp/server/http.py:297-306`). Without this the browser consent page — which POSTs same-origin — would be rejected.

## Breaking

The identity-provider redirect URI becomes `{origin}/mcp/auth/callback` and must be re-registered. Both the user and deployment guides are updated.

## Verification

- `bazel test //src/service/mcp/...` — 72/72 pass (incl. mypy, pylint)
- `bash deployments/charts/service/tests/render-tests.sh` — pass
- `bash deployments/charts/service/ci/validate-mcp-chart.sh` — pass, with route-scoped assertions that were negative-tested (removing the rewrite fails the check)

## Summary by CodeRabbit

* **New Features**
  * Added configurable browser-origin allowlisting for MCP connections.
  * MCP traffic and OAuth endpoints now consistently use the `/mcp` path.
  * Improved MCP proxy routing, metadata handling, and protected health checks.

* **Bug Fixes**
  * Corrected OAuth callback, issuer, and authorization URLs to use the MCP path.
  * Improved validation and forwarding of MCP request paths and origins.

* **Documentation**
  * Updated deployment and getting-started guides with the `/mcp` callback and metadata URLs.

## Checklist

### DIFF
--- a/deployments/charts/service/ci/validate-mcp-chart.sh
+++ b/deployments/charts/service/ci/validate-mcp-chart.sh
@@ -180,45 +180,43 @@ for expected in \
   assert_file_contains "$PROXY_MCP_MANIFEST" "$expected"
 done
 
+# FastMCP's OAuth endpoints are published under /mcp, so the gateway needs one
+# prefix route rather than an entry per endpoint name.
 for route in \
     mcp-protected-resource-metadata \
-    mcp-oauth-authorization-server-metadata \
-    mcp-oauth-authorize-get \
-    mcp-oauth-authorize-post \
-    mcp-oauth-oidc-callback \
-    mcp-oauth-register \
-    mcp-oauth-token \
-    mcp-oauth-consent-get \
-    mcp-oauth-consent-post \
-    mcp-oauth-authorization-server-metadata-options \
-    mcp-oauth-authorize-options \
-    mcp-oauth-oidc-callback-options \
-    mcp-oauth-register-options \
-    mcp-oauth-token-options \
-    mcp-oauth-consent-options; do
+    mcp-oauth \
+    mcp-authorization-server-metadata; do
   assert_route_contains "$PROXY_RENDERED_MANIFEST" "$route" 'cluster: osmo-mcp'
   assert_route_contains "$PROXY_RENDERED_MANIFEST" "$route" 'envoy.filters.http.jwt_authn:'
   assert_route_contains "$PROXY_RENDERED_MANIFEST" "$route" 'envoy.filters.http.ext_authz:'
 done
+# Scope the rewrite assertions to their routes: a file-wide grep for
+# 'prefix_rewrite: /' also matches the well-known route's longer value.
+assert_route_contains "$PROXY_RENDERED_MANIFEST" mcp-oauth 'prefix: /mcp/'
+assert_route_contains "$PROXY_RENDERED_MANIFEST" mcp-oauth 'prefix_rewrite: /'
+assert_route_contains "$PROXY_RENDERED_MANIFEST" mcp-authorization-server-metadata \
+  'path: /.well-known/oauth-authorization-server/mcp'
+assert_route_contains "$PROXY_RENDERED_MANIFEST" mcp-authorization-server-metadata \
+  'prefix_rewrite: /.well-known/oauth-authorization-server'
+
+# The health carve-out must answer 404 itself. Without the auth filters
+# disabled, jwt_authn answers 401 first and the carve-out is dead code that
+# only looks like it works.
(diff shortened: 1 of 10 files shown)


jiaenren:
## Runtime verification of the rewrite

Ran the exact route blocks from this PR through real Envoy (v1.31) against an echo upstream, to confirm the advertised URLs land on the paths the MCP SDK actually registers.

| Public URL | Upstream received |
|---|---|
| `/mcp` | `/mcp` |
| `/mcp/authorize` | `/authorize` |
| `/mcp/token` | `/token` |
| `/mcp/register` | `/register` |
| `/mcp/consent` | `/consent` |
| `/mcp/auth/callback` | `/auth/callback` |
| `/.well-known/oauth-authorization-server/mcp` | `/.well-known/oauth-authorization-server` |
| `/mcp/health` | **404** |
| `/mcp/health/live` | **404** |

This closes the loop with the unit test: `test_auth.py` asserts FastMCP advertises `https://osmo.example/mcp/authorize` while registering the handler at `/authorize`, and the table above shows the gateway bridging exactly that gap. The health carve-out also holds for nested paths, not just the exact `/mcp/health`.

jiaenren:
## Fixed: the health carve-out never actually fired

Found during a deployment run on a development instance.

`mcp-health-not-public` was the only MCP route without `typed_per_filter_config`, so `jwt_authn` returned **401 before `direct_response: 404` could run**. `/mcp/health` stayed protected, but only incidentally — by a different filter than the one meant to protect it. The carve-out was dead code that looked like it worked, and its own comment told future routes to copy the pattern.

The route now disables the auth filters so it owns its rejection:

```yaml
- name: mcp-health-not-public
  match: {prefix: /mcp/health}
  direct_response: {status: 404}
  typed_per_filter_config:
    {{- include "osmo.gateway-auth-filters-disabled" . | nindent 20 }}
```

Verified 404/404 on the dev instance with no regression elsewhere.

### Why the earlier verification missed it

The local Envoy harness I ran had **no `jwt_authn` filter at all**, so it could not reproduce the production filter ordering that caused this. It proved the rewrite semantics and nothing about filter precedence.

The durable fix is a render-time assertion rather than a better harness — `validate-mcp-chart.sh` now requires the health route to carry both filter-disable entries. I negative-tested it by removing the `typed_per_filter_config` again:

```
MCP chart validation failed: Gateway route mcp-health-not-public is missing: envoy.filters.http.jwt_authn:
```

The stack (#1329, #1330, #1331) has been rebased onto this fix.

RyaliNvidia (approved):

cypres (approved):

PR #1330, opened 2026-08-26
Take the MCP token-verification JWKS URI from OIDC discovery

jiaenren:
Stacked on #1329. Third of the MCP authentication simplification stack.

Issue #None

## Problem

Deployers supplied `accessTokenJwksUrl` — a value every OIDC discovery document publishes as `jwks_uri`, and which FastMCP already fetches at startup. OSMO hand-built a `JWTVerifier` purely to pin that URL, and carried a process-lifetime `httpx` client to serve it.

## Change

Let `OIDCProxy` build the verifier (`oidc_proxy.py:381-395`) and override only what discovery cannot supply. An Entra resource application configured for v1 access tokens issues them from `https://sts.windows.net/<tenant>/` even when its discovery document advertises the v2.0 issuer, so `accessTokenIssuer` stays configured — the JWKS URI now comes from discovery.

**Required OIDC-proxy values: 7 → 6.**

## One subtlety worth reviewer attention

The audience is applied *inside* the override rather than through `OIDCProxy`'s own `audience` argument. That argument is forwarded to the provider's authorize **and token** endpoints (`oidc_proxy.py:432-434`) — an Auth0-style parameter that Entra does not accept.

I got this wrong first. An existing assertion that no extra token parameters are sent caught it:

```
AssertionError: {'audience': 'https://osmo.example/mcp'} != {}
```

That test was written to prevent exactly this class of mistake, and it worked.

## Also removed

The `http_close_fails` test, which existed solely to verify the `try/finally` ordering between the httpx client and Redis — both the client and the ordering are gone.

## Verification

- `bazel test //src/service/mcp/...` — 72/72 pass, including a new assertion that the verifier's `jwks_uri` comes from the mocked discovery document while its `issuer` stays the configured one

## Checklist

### DIFF
--- a/deployments/charts/service/README.md
+++ b/deployments/charts/service/README.md
@@ -105,13 +105,10 @@ destination, but it cannot validate external DNS.
 | `services.mcp.extraVolumeMounts` | Additional MCP container volume mounts, including Vault-injected credential files. | `[]` |
 | `services.mcp.extraVolumes` | Additional MCP pod volumes. | `[]` |
 | `services.mcp.oidcProxy.enabled` | Enable FastMCP's built-in OIDC proxy inside the existing MCP process. It advertises CIMD and retains DCR as a compatibility fallback. | `false` |
-| `services.mcp.oidcProxy.scope` | Full delegated scope URI advertised to MCP clients and requested upstream, normally `<resourceUrl>/access_as_user`. | `""` |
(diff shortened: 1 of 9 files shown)


jiaenren:
Consolidated into #1329 to reduce the number of PRs in this stack for review.

Deriving the JWKS URI from OIDC discovery is part of the same theme: stop asking the deployer for values the deployment already knows or can discover.

PR #1331, opened 2026-08-26
Share OSMO's Redis for MCP proxy state and allow scaling out

jiaenren:
Stacked on #1330. Fourth of the MCP authentication simplification stack.

Issue #None

## Two fixes

**1. The single-replica rule was based on a misreading of FastMCP.**

The chart hard-failed on `replicas != 1` with the OIDC proxy enabled, citing process-local refresh serialization. FastMCP says the opposite:

> All state is stored in the configured `client_storage` backend (Redis, disk, etc.) **enabling horizontal scaling across multiple instances**
> — `oauth_proxy/proxy.py:212-213`

and `:1903-1912` handles the distributed refresh race explicitly ("another worker may have already refreshed and rotated the token"). The rule was also self-contradictory: sharing state through Redis only matters across replicas the chart forbade from existing.

**2. `oidcProxy.redis.tlsEnabled` did not inherit — and that one can CrashLoop the pod.**

`serviceName` and `port` already fell back to `services.redis`, but `tlsEnabled` did not. A deployment whose Redis requires TLS could satisfy every other OSMO service and still fail on MCP, with a connection error rather than a config error. All three now come from `services.redis`.

Only `dbNumber` stays MCP-local, which is its actual job — isolating proxy state from OSMO's other Redis users. Its default moves off `0` for the same reason.

## Test change

The negative test asserting the replica ban is replaced by a positive assertion that two replicas render. I negative-tested the replacement by reintroducing a ban and confirming the check fails:

```
MCP chart validation failed: MCP OIDC proxy with two replicas failed to render
```

## Verification

- `bazel test //src/service/mcp/...` — 72/72 pass
- Rendered Redis URL inherits correctly: `rediss://redis:6379/14`

## Checklist

### DIFF
--- a/deployments/charts/service/README.md
+++ b/deployments/charts/service/README.md
@@ -100,7 +100,7 @@ destination, but it cannot validate external DNS.
 | `services.mcp.scopes` | OAuth scopes advertised in direct-provider mode; ignored when `oidcProxy.enabled` is true. | `[]` |
 | `services.mcp.allowedOrigins` | Exact browser origins permitted on `/mcp`; native clients normally omit `Origin`. | `[]` |
 | `services.mcp.requestTimeoutSeconds` | Total timeout for each MCP-initiated Gateway request, from 1 through 60 seconds. | `10` |
-| `services.mcp.replicas` | Number of MCP replicas. Must remain `1` with `oidcProxy.enabled` because FastMCP 3.4.7 refresh serialization is process-local. | `1` |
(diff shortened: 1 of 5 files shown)


jiaenren:
Consolidated into #1329 to reduce the number of PRs in this stack for review.

Sharing OSMO's Redis and allowing replicas is the same theme: MCP inherits what every sibling service already inherits.

PR #1311, opened 2026-08-17
Add MCP task list tool

ethany-nv:
## Description

Adds the read-only `osmo_list_tasks` MCP tool for node-targeted on-call queries. It maps the existing `GET /api/task` filters, projects workflow/task/owner details, and preserves caller pool scope and pagination.

Issue - None

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/OSMO/blob/main/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

## Summary by CodeRabbit

* **New Features**
  * Added task listing through the MCP interface.
  * View tasks for a specific node with optional status, priority, and user filters.
  * Results support pagination and include task status, ownership, timing, retry, and placement details.
  * Task listings reflect accessible pools and support viewing tasks across users when permitted.

* **Bug Fixes**
  * Improved validation and handling of invalid inputs, empty results, and malformed responses.

PR #1333, opened 2026-08-27
Make the MCP smoke checks work in both authentication modes

jiaenren:
Stacked on #1331. Found by the first OIDC-proxy deployment.

Issue #None

## What the deploy found

The suite failed **before reaching any assertion**. It authenticates with the OSMO-issued OETF token, which FastMCP rejects — in proxy mode FastMCP verifies caller tokens against the identity provider. Evidence from the run: `POST /mcp` returned 200 for proxy-issued tokens and 401 for the suite's token within the same minute.

So this was never "a stale expectation". The suite was written for direct mode, where the Gateway authenticates and MCP trusts injected headers. Staging still runs direct mode, which is why the incompatibility only appeared when proxy mode was deployed for the first time.

**This matters for the PR that removes direct mode.** Once that lands, staging moves to proxy mode and this suite would break there too — so this fix is a prerequisite, not cleanup.

## Change

Split along the line that actually matters — whether a caller token is needed.

**Unauthenticated, runs in both modes** (new `test_public_discovery_surface`): the 401 challenge and its `resource_metadata` pointer, the RFC 9728 document, and in proxy mode that FastMCP advertises `authorization_endpoint` under `/mcp` and that the health endpoints the `/mcp` prefix would otherwise publish return 404.

This is the surface a client uses to *bootstrap* OAuth, so needing no token is the point.

**Authenticated** (catalog + workflow round-trip): skip in proxy mode with a message naming `OSMO_MCP_ACCESS_TOKEN`, which runs them there when an identity-provider token is supplied.

## Two related fixes

- The expected catalog was missing `osmo_list_tasks` (added by #1311), so the exact-equality assertion would have failed 25-vs-26 had it ever run. Two counts in `src/service/mcp/README.md` were stale for the same reason — corrected to 26 tools / 15 read-only, verified against the registry.
- **The suite never ran anywhere.** It carried an `auth` tag, and the kind environment's `exclude_tags: [auth]` dropped it — which is exactly why the staleness went unnoticed. It is now tagged by what it needs *deployed* (`mcp`), with kind excluding `mcp` explicitly, since kind keeps MCP disabled.

## Verification

- The behavioural half needs a redeploy of a development deployment to confirm against a live proxy-mode deployment

## Checklist

### DIFF
--- a/src/service/mcp/README.md
+++ b/src/service/mcp/README.md
@@ -268,7 +268,7 @@ provider. The MCP tool catalog and API-specific authorization do not change.
 
 ## Available tools
 
-The external catalog contains 25 tools: 14 read-only tools for caller-bound
+The external catalog contains 26 tools: 15 read-only tools for caller-bound
 health, profile, pool, resource, workflow, application, and
 credential-metadata inspection, plus four workflow actions, one
 profile-setting action, one credential action, four app lifecycle actions,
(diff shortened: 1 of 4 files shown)


jiaenren:
Consolidated into #1334 to reduce the number of PRs in this stack for review.

The smoke-check changes exist because direct mode is going away, so they belong with the commit that removes it.

PR #1334, opened 2026-08-27
Remove MCP direct identity-provider mode

jiaenren:
Stacked on #1333. The last structural PR in the MCP authentication simplification stack.

Issue #None

**+251 / -1,285.**

## Why

MCP shipped two ways to authenticate: the Gateway validating a bearer and MCP trusting injected headers, or FastMCP verifying the caller itself. Only one can be the answer, and the published docs already recommended the second. Every route, matcher, config field and code path was conditional on which one was active.

## Change

`request_context.py` **355 → 194 lines**. The ASGI middleware, per-request state, header parsing, and the `user_name` field no consumer read are gone. `get_request_credentials()` collapses to FastMCP's access token — so caller identity can no longer come from a header a caller could set. That is now structural rather than validated.

`auth_enabled` goes with it. Authentication isn't a mode, so its configuration is required rather than conditionally required: a deployment that cannot reach an identity provider fails at load instead of starting unauthenticated.

The Gateway stops synthesising what FastMCP serves — the protected-resource route forwards unconditionally, and the `local_reply_config` that rewrote the `/mcp` 401 is gone because FastMCP emits it. `authorizationServers`, `scopes`, and `oidcProxy.enabled` (which could only ever be true) are removed, along with the direct-mode CI fixture and the validator half that tested it.

Tests authenticate the way the service does: `protocol_harness` gains `AnyTokenVerifier` and a `service_config` helper, so suites exercising the tool pipeline get a real auth provider instead of trusted headers.

## Two behaviour changes, recorded not hidden

**1. Unauthenticated requests are rejected after the body is buffered, not before.** The deleted middleware enforced fail-before-body; FastMCP's auth runs inside the route, so `RequestBodyLimitMiddleware` — which is outermost — buffers the bounded body first. The request is still rejected and the body still bounded; only the order changed. A test asserts this explicitly rather than leaving it undiscovered.

**2. After a cancelled request, a follow-up on the same client is FastMCP's to decide.** I verified against the pre-change tree, with an auth provider installed, that an authenticated deployment **already behaved this way** — so this PR does not introduce it. It is pre-existing in OIDC-proxy mode and worth its own investigation; a development deployment has it today.

## Verification

- User guide, deployment guide, appendix, service README and chart README no longer document a mode that does not exist

## Checklist


jiaenren:
Superseded by #1350, which is the same branch and the same content with the smoke-check commit from #1333 folded in.

GitHub refused to retarget this PR's base (`Cannot change the base branch because the pull request is part of a stack`) after the stack was consolidated from 12 PRs to 6, so it had to be reopened rather than retargeted. No content changed.

PR #1335, opened 2026-08-27
Fold MCP chart validation into the standard render tests

jiaenren:
Issue #None

**+87 / −278.**

## Why

MCP was the only chart area in the repo with a dedicated `ci/validate-*.sh`. It had grown to 272 lines against a house harness (`tests/render-tests.sh`) of 78 that covered MCP not at all, and it ran from its own CI step — so MCP chart rendering was checked by a mechanism nothing else uses.

Most of what it asserted wasn't behaviour. Numeric ranges, URL shapes and scope patterns are enforced by `MCPAuthConfig` at start-up; re-proving them at render time stated the same contract twice in two languages, and every new bound needed editing in both places.

## What moved

The behavioural assertions, in the house harness's own style:

- MCP renders nothing when disabled
- the gateway publishes one `/mcp` prefix route and rewrites it off
- **the health carve-out answers 404 itself** rather than leaning on `jwt_authn` — the defect your deploy run found
- Redis is inherited from `services.redis`
- derived values do not reappear as deployer inputs
- the proxy scales out
- `extraEnv` cannot redirect the relay, and a malformed `resourceUrl` is rejected

The roles Lua filter assertions came along because they are its **only coverage anywhere in the repo**.

## Each assertion was negative-tested

Not just written — I reintroduced the defect each one guards and confirmed the harness fails:

```
health carve-out: caught
prefix_rewrite:   caught
```

## Before / after

```
ci/validate-mcp-chart.sh:   272 lines → deleted
tests/render-tests.sh:       78 → 163 lines
total chart-test lines:     350 → 163
dedicated MCP CI step:      yes → none (uses the same matrix as every other chart)
```

## Also

The proxy fixture moves to `tests/` alongside the harness that uses it. `ci/` is now empty, and the dedicated CI step in `helm-chart-lint.yaml` is gone — MCP is now checked by the same `render-tests.sh` step as every other chart.

## Verification

- `bash deployments/charts/service/tests/render-tests.sh` — pass, now including MCP

## Checklist


jiaenren:

The render assertions for the reduced configuration contract live in the harness this PR creates, and in the proxy fixture it renames, so it cannot be separated from the change it validates.

PR #1343, opened 2026-08-27
Resolve the MCP signing key

jiaenren:
Stacked on #1342. Resolves the signing key, the one step of the simplification that had no PR, plus the key-derivation change #1331 left behind.

Issue #None

## Enforce the entropy the derived keys depend on

Both the proxy-token signing key and the Redis storage key are derived from the client secret, so its entropy is theirs, and nothing checked it. The service now refuses to start below 32 characters.

Worth flagging: the deployment runbook already told operators the service refused to start below that length, by design. It did not. An operator reading that reasonably believed a guard existed. Now one does.

## Stop importing a private symbol

`derive_jwt_key` came from `fastmcp.server.auth.jwt_issuer`: private, with no deprecation contract. A minor release could drop it, which is loud, or change its derivation, which is quiet and much worse, since every stored registration and token would become undecryptable.

The derivation is now local with the same HKDF parameters (`SHA256`, length 32, salt, `info=b"Fernet"`, urlsafe-base64), so **the bytes are identical and existing state stays readable**. `test_auth` asserts that equivalence against FastMCP directly, so a change on their side surfaces as a test failure rather than as unreadable Redis.

## Replace the production-readiness warning

The deployment guide said the chart exposes no alternative to the derived key and told operators to assess the limitation before rolling out, text that cannot ship with an official release.

That framing does not survive reading FastMCP 3.4.7: `proxy.py:487-495` passes the client secret as `high_entropy_material` through HKDF, while the PBKDF2 path carrying a length warning (`:497-506`) is reserved for low-entropy operator-supplied strings. The property the derived key depends on is client-secret entropy, which is now stated and enforced.

`jwtSigningKey` is deliberately **not** added. With the entropy check and the corrected documentation in place the derived key is defensible on FastMCP's own terms. Deployments requiring key separation can add it later without changing this contract.

## Verification

- `bazel test //src/service/mcp/... //test/smoke/...` passes 76/76, including a new test that a short secret fails at startup, and the storage-key test now proving byte equivalence with FastMCP
- `render-tests.sh` passes


jiaenren:

Resolving the signing key is part of the same theme: settling what a deployer must supply and what the service derives or validates for itself.

PR #1344, opened 2026-08-27
Retire the mcp:Access action

jiaenren:
Stacked on #1343. Retires an RBAC action that is never evaluated.

Issue #None

**+2 / -100.**

## Why

`mcp:Access` mapped to exactly one path, `/mcp`, and `ext_authz` is disabled on that route, so it was never evaluated. A role granting it implied an access control that did not run, which is worse than not having the action at all.

The wholesale gate is the identity-provider scope: `JWTVerifier` rejects a token without it, so a caller cannot reach the protocol endpoint. Per-operation authorization is unchanged, because every tool call reaches OSMO through `/api`, where gateway JWT validation and semantic RBAC apply exactly as they do for the CLI and UI.

`resourceTypeMCP` and `ResourceTypeMCP` go with it; nothing outside the registry referenced either. The two removed tests existed only to assert this action's path mapping.

## No migration needed

`IsValidAction` has no callers, so nothing validates stored role actions against the registry. A role row still granting `mcp:Access` simply keeps an action matching no path, which is already the case today since the route bypasses `ext_authz`. Removing it from the seeded default role means the grant stops being handed out; it does not need clearing from existing rows.

## Verification

- `bazel test //src/utils/roles/...` passes 2/2, including the integration test
- `bazel test //src/utils/connectors/tests:test_default_roles` passes
- `render-tests.sh` passes

## Left alone

`deployments/charts/osmo/tests/control-mcp-values.yaml` also contains the string `mcp:Access`, but under `scopes:`. That is an OAuth scope in the direct-mode fixture, a different concept that happens to share the spelling. It belongs with the direct-mode removal, not here.


jiaenren:

Retiring `mcp:Access` removes another thing a deployment had to declare that never took effect.

PR #1347, opened 2026-08-28
Clean up the MCP authentication simplification

jiaenren:
Last PR in the MCP authentication simplification stack.

Issue - None

A cleanup pass over the whole MCP auth stack, from a four-angle review (reuse, simplification, efficiency, altitude). **No behaviour changes** — the chart renders byte-identically with and without MCP enabled.

## Dead state the simplification left behind

`$mcpOidcProxyEnabled` was an alias for `$mcpEnabled`, so its three `{{- if }}` blocks were nested inside the very condition they tested. `$mcpMetadataUrl` and `$mcpBaseUrl` lost their only reader when the `www-authenticate` mapper was deleted. `mcp-service.yaml` pre-declared six variables to hoist scope out of a conditional that no longer exists.

## Credentials derived once per tool call instead of three times

Deleting `RequestContextMiddleware` left `get_request_credentials()` recomputing from scratch on every call — twice in `protocol.py`, then again per Gateway request in `tool_requests.py`. Each derivation is a regex over the bearer plus a scan of the request-ID header, on the request hot path. `track_request_task()` had become a contextmanager whose body was a bare `yield`; it is now `bind_credentials()`, which binds one derivation so later calls reuse it.

## Mandatory configuration stops pretending to be optional

The five fields auth cannot start without were `Optional`-with-`None` plus a hand-rolled missing-field check — which then cost eight `cast(str, ...)` calls to undo. Declaring them required lets pydantic report them by name, and mypy now catches an incomplete construction at build time.

## Tests

- `AnyTokenVerifier` re-implemented FastMCP's shipped `DebugTokenVerifier`; the `user_name` threaded through the harness only ever wrote a claim nothing reads.
- `test_auth` passed `auth_enabled` and `oidc_access_token_jwks_url` — both deleted by this stack and silently ignored, one contradicting an assertion five lines above it.
- `test_protocol` hand-built its ASGI app, so it skipped the origin guard and body limit `create_application` installs.
- Two `create_autospec(Redis)` calls cost ~1s to assert one awaited close.
- Three BUILD targets kept a dep on `//src/lib/utils:login` they no longer use.

## Render harness

It rendered the same release twice to slice one template out, and counted the MCP audience across the whole document — passing only because the other occurrence happens to carry a `value:` prefix. Now scoped to the `audiences` list, and negative-tested.

## Docs

Deleting direct mode took four sections of the user appendix with it that were never about auth modes, orphaning the Safety and Limits bullets under an authentication lead-in. Restored. The deployment guide still showed `oidcProxy.enabled` and the secret paths the chart now derives; the README still told operators to keep `replicas` at `1`, which the move to Redis had already made false.

Comments are shortened throughout. `_derive_fernet_key` no longer calls `derive_jwt_key` private — it isn't; the real reason for pinning the derivation locally is that an upstream change would silently make stored state undecryptable.

## Before / after

```
auth.py:                            360 → 321 lines
Envoy template:                   1,219 → 1,196 lines
mcp-service.yaml placeholders:        9 →   3
test/smoke/mcp_checks.py:           415 → 382 lines
credential derivations per tool call: 3 →   1
```

Everything above is behaviour-preserving: the rendered chart is byte-identical
with and without MCP enabled.

## Verification

- `render-tests.sh` — pass; MCP and non-MCP renders byte-identical to before
- Negative-tested the audience assertion by making the append unconditional and confirming the harness fails

## Checklist


jiaenren:

Every line of this cleanup fixed something the other three introduced, so it had no home of its own; its content is distributed into #1350, and the umbrella-chart portion into #1360.

PR #1366, opened 2026-09-03
Show clear registry errors when submitting workflows

jiaenren:
## Summary

Report missing images, invalid credentials, registry throttling, and outages as distinct submission errors. Failed registry validation returns the saved workflow ID; the CLI shows a useful message even when error fields are empty.

MCP submission, app submission, validation, and restart expose allowlisted error codes while keeping upstream messages private.

## Testing

- Unit and local registry integration tests cover classification, credential fallback, workflow persistence, and CLI text/JSON errors.
- MCP protocol tests cover all four submission-related tools, including error codes, redaction, and no retries. Focused Bazel tests, type checks, and lint passed.
- Dev checks passed through CLI/API and authenticated MCP submission (`1ea7da098`): API/MCP return `IMAGE_NOT_FOUND`, failed workflows are queryable, and valid-digest workflows complete with expected logs.
- Validation/restart follow-up is verified by protocol tests and live dev instance.

Issue - None

## Checklist


jiaenren:
> Update `registry_auth` so an initial unauthenticated 404 proceeds through the credentialed authentication flow instead of returning immediately

Good catch — fixed in 8e3b51a4d.

The gap was real. The initial `HEAD` in `registry_auth` is unauthenticated (credentials are only used later, in the token exchange), so returning on any non-401 meant a concealed-private-repository 404 short-circuited before a credential was ever tried. Every credentialed retry from `validate_registry` just repeated the same anonymous 404, and the comment I left there claiming credentials were still tried was wrong.

Rather than special-casing 404, the token exchange is now driven by the `WWW-Authenticate` challenge, which is what the registry auth spec actually keys on:

```python
if response.status_code == 200 or 'www-authenticate' not in response.headers:
    return response
```

A 404 carrying a challenge now runs the credentialed token exchange; a 404 without one is still returned for classification as an unresolvable tag. This also preserves 401 Bearer handling unchanged, and incidentally turns a malformed 401 with no challenge header into a clean credential error instead of a `KeyError`.

Regression test added in `test_common.py`: `test_concealed_private_repository_404_authenticates_with_credentials` asserts the concealed 404 drives both the token request and the credentialed manifest fetch, and returns the 200. It fails against the previous code (the concealed 404 is returned directly). The existing no-challenge case is retained as `test_404_without_an_auth_challenge_is_returned_for_the_caller_to_classify`.

jiaenren:
Verification note for reviewers, on where this has and has not been exercised.

Covered:
- `//src/service/core/workflow/tests:test_workflow_service` submits against a real registry container: an unresolvable tag returns 400 with `error_code: IMAGE_NOT_FOUND` and a populated `workflow_id`, and the same repository submitted by digest still returns 200. Reverting the fix turns that into `'CREDENTIAL' != 'IMAGE_NOT_FOUND'` with `workflow_id: None`, which is the reported production symptom.
- The Local KIND Deployment job deploys OSMO and runs the kind smoke suite green on this branch.
- Unit coverage for every status mapping, transport timeouts and connection errors, the credential-loop short-circuit, the concealed-404 challenge path, and the client rendering the service body instead of an empty message.

Not covered: I attempted a deploy to an internal dev instance for a live check against the Docker Hub image from the original report. The image built and pushed and the deployment tag was updated, but the cluster's continuous-delivery sync never picked it up within its 30 minute budget, which is an infrastructure issue unrelated to this change. That live check was not completed.

jiaenren:
Verified the reported image against live Docker Hub today, through the real code path rather than a mock. The registry still behaves exactly as reported:

```
GET /v2/hijkzzz/molt/manifests/0.1
-> 404 {"errors":[{"code":"MANIFEST_UNKNOWN","message":"manifest unknown","detail":"unknown tag=0.1"}]}

GET /v2/hijkzzz/molt/manifests/sha256:f3fd67367cd07331b1057096018c6fb12a58d421e43dc6faff22217d54200e6d
-> 200 OK
```

Driving `registry_auth` + `registry_manifest_error` against that live response:

```
LIVE tag hijkzzz/molt:0.1 -> HTTP 404
LIVE error type: OSMOImageNotFoundError
LIVE error_code: IMAGE_NOT_FOUND
LIVE message: Could not resolve image tag hijkzzz/molt:0.1: registry-1.docker.io returned 404
              MANIFEST_UNKNOWN. Verify that the tag exists or submit an immutable digest.
LIVE digest -> HTTP 200
```

That message matches the wording the issue asked for, and the same repository still resolves by digest, so the tag-404 vs digest-200 split is confirmed against the real registry and not just the registry container used in `test_workflow_service`.

The check itself was a throwaway harness run locally against the live registry; it is deliberately not part of this PR, since a test that depends on a third-party tag staying absent would be flaky in CI. The committed coverage pins the same behavior against a registry container.

jiaenren:
Verified end to end on a live OSMO deployment, submitting the image from the original report through the real submit API. Same command, same cluster, same pool — only the service build differs.

Before (deployment running this branch's merge base):

```
$ osmo workflow submit bad-tag.yaml --pool <pool> ...
Server responded with status code 400
Error message: Workflow submit failed:
Unable to authenticate for pulling image hijkzzz/molt:0.1. Please create a credential
matching registry-1.docker.io/hijkzzz/molt or check if the image exists.
```

After (same deployment, service image built from this branch):

```
$ osmo workflow submit bad-tag.yaml --pool <pool> ...
Server responded with status code 400
Error message: Could not resolve image tag hijkzzz/molt:0.1: registry-1.docker.io returned
404 MANIFEST_UNKNOWN. Verify that the tag exists or submit an immutable digest.
```

And the same repository by immutable digest, on the same build, submits normally:

```
$ osmo workflow submit digest.yaml --pool <pool> ...
Workflow submit successful.
Workflow ID        - osmo6683_digesttest-1
```

So the tag-404 vs digest-200 split from the report is confirmed against a real deployment and the live Docker Hub registry, not only against the registry container in `test_workflow_service`. The test workflow was cancelled afterwards and the environment returned to its original build.

One caveat on scope: the CLI used for this check was a pre-existing release build, so it exercised the service-side classification only. The client-side changes in this PR (empty-message fallback, `workflow_id` preservation, registry error typing) are covered by the unit tests in `test_client.py` rather than by this run. Worth noting that the old client still rendered the new message correctly, since it falls through to `OSMOUserError` for the unrecognized `IMAGE_NOT_FOUND` code.

jiaenren:
Thanks — all four are real. Fixed in 3ab7d77b3.

**1. Anonymous 429 must allow authentication (P1).** You're right, and this was the worst kind of mistake to make on this branch: I reasoned "no credential can fix a rate limit," which is false for the case that actually happens. Docker Hub refuses anonymous pulls on quota and serves the same pull once authenticated, so treating a 429 as terminal before any credential reported a fixable condition as unfixable — the exact bug class this PR exists to remove. A 429 now only ends the sequence once a credential has already been used; the predicate is renamed `registry_failure_needs_backoff` and documents why anonymous 429s are excluded. Covered by `test_validate_registry_authenticates_through_an_anonymous_rate_limit`.

**2. Stop retrying 404s credentials cannot resolve.** `registry_auth` now returns a `RegistryAttempt` carrying whether the registry offered a `WWW-Authenticate` challenge, which is the only condition under which a credential can change the answer. Without one the loop is skipped entirely, so the unchallenged 404 is requested once rather than once per credential. Separately, a definitive 404 from the anonymous attempt is preserved instead of being overwritten by a later credential's 401 — that overwrite turned a correct "tag does not exist" back into an authentication error, which is the original complaint in miniature. Covered by `test_validate_registry_does_not_repeat_an_unauthenticatable_request` and `test_validate_registry_keeps_a_missing_manifest_over_a_later_rejection`.

**3. Check outages before following a challenge.** A 5xx now returns before the challenge is considered, so an outage no longer spends a token exchange and a manifest GET only to surface as a credential error. Covered by `test_server_error_with_a_challenge_does_not_start_a_token_exchange`, which asserts no follow-up request is issued.

**4. Distinguish token-service failures from manifest failures.** `registry_auth` no longer returns arbitrary token-endpoint responses into manifest classification. A 401/403 is still returned, since that means the supplied credential was rejected and the caller should try the next one; anything else raises `OSMORegistryUnavailableError` naming the token endpoint and its status, so a token service 404 no longer tells users their tag is missing. Covered by `test_token_service_failure_is_not_reported_as_a_manifest_failure` and `test_rejected_credential_is_returned_so_the_caller_can_try_the_next`.

**On the simplification** — fetching the challenge once and reusing it across credential attempts: not taken, deliberately. The repeated *unchallenged* requests you measured are gone via the challenge gate above. What remains is that each credentialed attempt re-runs the `HEAD` inside `registry_auth`, which costs one extra request per credential on an already-failing submit. Splitting `registry_auth` into probe/fetch stages to remove it would change the shape of a function on the submit path with a second caller (`RegistryCredential.valid_cred`), and I'd rather not widen this change further for one request per credential. Happy to do it as a follow-up if you think it's worth it.

70 tests pass across the same targets, including the registry-container integration test.

cypres (dismissed):

tdewanNvidia (approved):

cypres (approved):

PR #1409, opened 2026-09-16
Fix data-transfer timeout deadlock

### LINKED ISSUE #1408: Data-transfer timeout leaves osmo-ctrl blocked on watchdog channel sends
### Description

`createOutCommandStream()` starts a watchdog goroutine that monitors stdout inactivity. The stream function creates an unbuffered `quit` channel and after stdout scanning finishes, sends `quit <- true` followed by `timeoutChan <- false`.

When inactivity reaches `DataTimeout`, the watchdog kills the subprocess, sends `true` to `timeoutChan` and exits. Once stdout closes, the stream function attempts to send to `quit`, but the watchdog that would receive that value has already exited. The send blocks indefinitely, preventing the deferred `waitStreamLogs.Done()` call.

`common.RunCommand()` waits for that wait group before consuming the timeout result. Consequently, the command never returns its timeout error and the surrounding retry loop cannot advance.

### Relevant code

- `src/runtime/pkg/data/data.go`: `createOutCommandStream()` and `RunOSMOCommandStreamingWithRetry()`.
- `src/runtime/pkg/common/common.go`: `RunCommand()`.
- `src/runtime/pkg/data/data_runtime_test.go`: existing stream tests cover ordinary scanner completion but do not exercise this timeout sequence.

### Reproduction sequence

1. Start a subprocess that remains alive without emitting stdout.
2. Allow stdout inactivity to exceed `DataTimeout`.
3. The watchdog kills the process and writes `true` to the buffered timeout channel.
4. The watchdog exits and the stdout scanner finishes.
5. The stream goroutine blocks sending to the unbuffered `quit` channel.
6. `RunCommand()` remains blocked in `waitStreamLogs.Wait()`.

### Impact

Affected download/upload subprocess timeouts can leave osmo-ctrl indefinitely stuck instead of retrying or reporting the timeout. External termination may be required to stop the stuck execution.

### Verification

An isolated reproduction used the existing stream function, a real sleeping subprocess and a shortened timeout. The goroutine dump showed the stream blocked on the channel send and the caller blocked in `WaitGroup.Wait()`, with one timeout result already queued.

### Suggested resolution

Use cancellation that does not require a live receiver, such as closing a completion channel or cancelling a context. Coordinate the watchdog and scanner so exactly one terminal timeout result is sent and the watchdog exits cleanly.

Merely buffering `quit` is insufficient: the subsequent `timeoutChan <- false` would block because the capacity-one channel already contains `true` and its receiver waits for stream completion first. Timeout reporting must also preserve the timeout outcome when scanner completion follows process termination.

Add a regression test that forces a real timeout and verifies bounded completion with a timeout error, alongside normal-completion coverage. Synchronize shared progress timestamps when revising the watchdog.

abhinavgautam01:
## Description

Fix a deadlock that prevents osmo-ctrl from retrying timed-out data transfers.

After killing an inactive subprocess, the watchdog previously exited before receiving the stdout stream's shutdown signal. The stream then blocked indefinitely, preventing `RunCommand()` from returning its timeout error.

Use a closeable stop channel and wait for the watchdog to finish before reporting exactly one timeout result. Protect the shared progress timestamp with a mutex and reap timed-out subprocesses before returning.

Add regression coverage for timeout completion, retry recovery, progress resetting the timeout, successful commands and nonzero exits.

Validation:
- Confirmed the timeout regression test fails before the fix and passes afterward.
- Passed Bazel data/common tests, including race-enabled runs.
- Passed `go vet`, `gofmt` verification and `git diff --check`.
- Built osmo-ctrl and osmo-user for Linux AMD64 and ARM64.

Closes #1408

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved command timeout handling so timed-out processes are fully completed and correctly reported.
  - Prevented inactive output streams from being incorrectly marked successful after exceeding the configured timeout.
  - Added progress-aware timeout behavior, allowing active commands to continue while terminating stalled commands and their child processes.
  - Improved reliability when retrying streaming operations after a timeout.

- **Tests**
  - Expanded coverage for command success, failures, empty output, timeouts, progress updates, process completion, and stream recovery.


abhinavgautam01:
The watchdog now remains active until the subprocess is reaped, including when stdout closes early. Timeout handling also terminates the data command’s process group. Added regression coverage for early output closure and retry recovery.

